PR #11, opened 2025-05-29
fix: Fixes for cycle marker and opcode stats

antoniolocascio:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


antoniolocascio:
@shamatar I can do that when we remove the old module. This new one already depends on the old one, so I opted for not duplicating more code.

antoniolocascio:
Ah I see, can we do that later? This is blocking merging PRs in zk os

PR #23, opened 2025-05-31
feat: update linker scripts to put stack on top of ROM for auto-underflows

shamatar:
## What ❔

Also limit HEAP size to 768Mb, and put a stab to eventually allow initialization of .data section (so we can remove few `MaybeUnints` in `static mut`). Also by default we can same prover RAM by allocating and tracking less memory per simulator (1Gb only instead of 4Gb)

## Why ❔

Current ROM/RAM access piece of circuits will catch stack underflows for free

## Is this a breaking change?
- [x] Yes
- [ ] No

NOTE: linker scripts need to be propagated to OS repo, and insanely large blocks can fail with OOM at allocation due to reduced heap size

## Checklist


shamatar:
I also rebuilt all old binaries just in case

shamatar:
@mm-zk CI failed to find a second proof file, because previously Fibonacci was different, and we were using old binary all the time. Now it was rebuilt, and fits into 1 circuit. Do you remember what was the very original one? We also do not assert the outputs, and it's now producing a different one

robik75 (approved):

mm-zk:
in the past, I've compiled 'basic fibonacci' in debug mode -- so that it consumed more circuit steps.

PR #38, opened 2025-06-06
fix(cs,gpu_prover): fix SRASignFiller table function

robik75:
## What ❔

Fix for a bug in the SRASignFiller table function


shamatar:
It should be shift amount == 0, not index. If CPU code uses index, then it's a bug by itself

robik75:
> It should be shift amount == 0, not index. If CPU code uses index, then it's a bug by itself

this? https://github.com/matter-labs/zksync-airbender/blob/main/cs/src/tables.rs#L1188

PR #48, opened 2025-06-14
feat: Batch inversion in quotient_evaluator

hatemosphere:
## What ❔

Optimize quotient evaluator with batch field inversions:

 - Collects all values to invert (4 per row) into a single batch
 - Performs one batch inversion operation instead of 4 individual inversions
 - Properly distributes inverted values back to their respective divisor computations

## Why ❔

Tackling the TODO. 
~5x average speedup for field inversion operations

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


0xVolosnikov (approved):
Generally LGTM. 

But please also get a review from @mm-zk or @shamatar since I'm not very deeply in the context here.

shamatar:
I look at the code and have a hard time to understand what it did before and why this way. Divisors are in the form of (X^N - 1)/(X - 1), but we only divide by them, so batch inversions are not even needed in that form - (X^N - 1) is constant, and instead it's easy to compute just (X - 1)/(X^N - 1) and multiply by them later. I'll check properly when back from vacation

PR #50, opened 2025-06-16
feat(gpu_prover): new GPU-backed execution prover

robik75:
## What ❔

This PR adds a new GPU-backed execution prover - in this case execution means a program binary execution.
This new prover encapsulates all the pieces needed to be able to produce proofs on potentially multiple GPUs in one function call.
Most important features:
- supports multithreaded proof calls, multiple execution proofs can be produced in parallel
- generates and caches precomputations for the configured binaries/circuit types
- no need to instantiate a host allocator, the prover calculates the amount of pinned memory need, creates the allocator and allocates pinned host memory based on the circuits needed for the configured binaries, the number of GPUs and the maximum number of concurrent proving tasks
- multi-treaded tracing with fast-forwarding and separate RAM/cycles/delegation tracing logic to produce the traces as fast as possible
- GPU workers with multi-stage execution to overlap transfers with scheduling and result retrieval
- GPU workers have a setup cache, proof of consecutive circuits of the same type can reuse the setup cached on the device
- GPU manager managing work queues, distributing work to GPU workers based on the state of their respective work queues, being able to push forward the pipelines if the work queue is running empty and also flush pipelines at the end of execution
- graceful shutdown of all the workers
- log logging facade is used throughout the code to plug it into the existing logger

## Why ❔

Because it has what plants crave!

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


gpu_prover/src/execution/prover.rs
```diff
@@ -0,0 +1,930 @@
...
+};
+use worker::Worker;
+
+type A = ConcurrentStaticHostAllocator;
+
+const CPU_WORKERS_COUNT: usize = 6;
+const CYCLES_TRACING_WORKERS_COUNT: usize = CPU_WORKERS_COUNT - 2;
+
+#[derive(Clone)]
+pub struct ExecutableBinary<K: Clone + Debug + Eq + Hash, B: Into<Box<[u32]>>> {
```
mm-zk, line 55: btw - please add comments - especially around what "key" is

PR #54, opened 2025-06-18
fix(gpu_prover): Ensure deterministic ordering of delegation data

mcarilli:
## What ❔

Ensures deterministic ordering of delegation data when unpacking it from HashMaps.

## Why ❔

In current code, some delegation data is unpacked directly from HashMaps, so the order is nondeterministic. However, other parts of the code expect it to be ordered by delegation type (id). For big proofs (e.g. Ethereum-block-equivalent) that use multiple delegation types, I sometimes saw [this assert](https://github.com/matter-labs/zksync-airbender/blob/main/circuit_defs/trace_and_split/src/lib.rs#L542) fire, indicating the wrong order. With these diffs, I no longer see the assert fire.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/gpu_prover/src/execution/prover.rs
+++ b/gpu_prover/src/execution/prover.rs
@@ -659,14 +659,29 @@ impl<K: Clone + Debug + Eq + Hash> ExecutionProver<'_, K> {
             .sorted_by_key(|(index, _)| *index)
             .map(|(_, caps)| caps)
             .collect_vec();
-        let delegation_memory_commitments = delegation_memory_commitments
-            .into_iter()
-            .map(|(id, caps)| (id as u32, caps))
+        let mut delegation_memory_commitment_keys =
+            delegation_memory_commitments.keys().copied().collect_vec();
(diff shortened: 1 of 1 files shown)


gpu_prover/src/execution/prover.rs
```diff
@@ -659,15 +659,30 @@ impl<K: Clone + Debug + Eq + Hash> ExecutionProver<'_, K> {
             .sorted_by_key(|(index, _)| *index)
             .map(|(_, caps)| caps)
             .collect_vec();
-        let delegation_memory_commitments = delegation_memory_commitments
-            .into_iter()
-            .map(|(id, caps)| (id as u32, caps))
+        let mut delegation_memory_commitment_keys =
```
mm-zk, line 662: please add comments in both changes, explaining that we must process them in correct order.

mm-zk (dismissed):

mm-zk (approved):

PR #55, opened 2025-06-20
Preparing branch for audit (do not merge)

Fitznik:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/cs/src/delegation/bigint_with_control/mod.rs
+++ b/cs/src/delegation/bigint_with_control/mod.rs
@@ -376,19 +376,6 @@ pub fn define_u256_ops_extended_control_delegation_circuit<F: PrimeField, CS: Ci
                     &subtraction_result_limbs[i].1,
                     &intermediate_carry_value,
                 );
-            // rest
-            value = <<CS as Circuit<F>>::WitnessPlacer as WitnessTypeSet<F>>::U16::select(
-                &assign_subtraction_result,
-                &subtraction_result_limbs[i].0,
-                &value,
-            );
-            intermediate_carry_value =
-                <<CS as Circuit<F>>::WitnessPlacer as WitnessTypeSet<F>>::Mask::select(
-                    &assign_subtraction_result,
-                    &subtraction_result_limbs[i].1,
-                    &intermediate_carry_value,
-                );
-
             placer.assign_u16(additive_ops_result[i], &value);
             placer.assign_mask(intermediate_carry_variables[i], &intermediate_carry_value);
         }
--- a/cs/src/devices/optimization_context.rs
+++ b/cs/src/devices/optimization_context.rs
@@ -117,10 +117,11 @@ pub struct OptimizationContext<F: PrimeField, C: Circuit<F>> {
 }
 
 impl<F: PrimeField, CS: Circuit<F>> OptimizationContext<F, CS> {
+    #[inline(always)]
     pub fn save_indexers(&self) -> OptCtxIndexers {
         self.indexers
     }
-
+    #[inline(always)]
     pub fn restore_indexers(&mut self, indexers: OptCtxIndexers) {
         self.indexers = indexers
     }
@@ -144,7 +145,7 @@ impl<F: PrimeField, CS: Circuit<F>> OptimizationContext<F, CS> {
             _marker: std::marker::PhantomData,
         }
     }
-
+    #[inline(always)]
     pub fn reset_indexers(&mut self) {
         self.indexers.reset()
     }
@@ -462,8 +463,6 @@ impl<F: PrimeField, CS: Circuit<F>> OptimizationContext<F, CS> {
             let mask = placer.get_boolean(exec_flag_var);
             witness_early_branch_if_possible(mask, placer, &evaluate_fn_inner);
         };
-
-        cs.set_values(value_fn);
 
         cs.set_values(value_fn);
 
--- a/cs/src/machine/machine_configurations/state_transition_parts/decode_and_read_operands.rs
+++ b/cs/src/machine/machine_configurations/state_transition_parts/decode_and_read_operands.rs
@@ -4,98 +4,6 @@ use crate::machine::{
     ops::{RD_STORE_LOCAL_TIMESTAMP, RS1_LOAD_LOCAL_TIMESTAMP, RS2_LOAD_LOCAL_TIMESTAMP},
 };
 
-pub(crate) fn optimized_decode_and_read_reg_operands<
-    F: PrimeField,
-    CS: Circuit<F>,
-    const ASSUME_TRUSTED_CODE: bool,
-    const PERFORM_DELEGATION: bool,
-    const ROM_ADDRESS_SPACE_SECOND_WORD_BITS: usize,
->(
-    cs: &mut CS,
-    pc: Register<F>,
-    opcodes_are_in_rom: bool,
-    decode_table_splitting: [usize; 2],
-) -> (
-    Vec<ShuffleRamMemQuery>,
-    RS2ShuffleRamQueryCandidate<F>,
-    Register<F>,
-    Register<F>,
-    Constraint<F>,
-    OptimizedDecoderOutput<F>,
-    Vec<Boolean>,
-) {
-    let next_opcode = read_opcode_from_rom::<F, CS, ROM_ADDRESS_SPACE_SECOND_WORD_BITS>(cs, pc);
-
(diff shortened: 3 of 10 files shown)


examples/big_int/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+
+
+unsafe fn workload() -> ! {
+    let mut a = U256([1, 1, 1, 1]);
+    let b = U256([1, 1, 1, 1]);
+    let mut round_mask = 1;
+    csr_trigger_delegation( a.0.as_ptr(), b.0.as_ptr(), round_mask );
+
+
+    zksync_os_finish_success(&[round_mask, 0, 0, 0, 0, 0, 0, 0]);
```
mcarilli, line 96: Imo we should output the result as well (`a[0], a[1], a[2], a[3]`) to sanity check what it did.

examples/big_int/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+#[export_name = "_machine_start_trap_rust"]
+pub extern "C" fn machine_start_trap_rust(_trap_frame: *mut MachineTrapFrame) -> usize {
+    {
+        unsafe { core::hint::unreachable_unchecked() }
+    }
+}
+
+#[inline(always)]
+fn csr_trigger_delegation(
+    input_a: *const u32,
```
mcarilli, line 66: The circuit's memory manipulation will probably technically work if `input_a` is `const` or `mut`, but it should be mut to signal the intention, like https://github.com/matter-labs/zksync-os/blob/main/crypto/src/bigint_delegation/delegation.rs#L98

antoniolocascio:
Cannot really comment on files that have no diff, but https://github.com/matter-labs/zksync-airbender/blob/audit_branch/docs/tutorial.md might need a bit of polishing and updating. For example, it says we only have Blake precompile.

It would be nice to have link to other doc pages that expand on the different aspects.

antoniolocascio:
This link is broken: https://github.com/matter-labs/zksync-airbender/blob/audit_branch/docs/custom_assumptions.md
Did you remove this section or is it missing?

antoniolocascio:
The docs need more structure, which should be reflected in the README. Right now the index is mixing theoretical descriptions, code docs and running tutorials. These 3 should be separated.

antoniolocascio:
Is https://github.com/matter-labs/zksync-airbender/blob/audit_branch/docs/repo_layout.md up to date? Is this stable? Sometimes keeping such a tight coupling of code structure and docs might become a headache, as it's bound to get outdated. Personally, I would try to simplify this page (just keep descriptions for important/stable crates/code directories) and make sure all of this documentation is also part of the code, where it's more likely to be updated if code is changed

docs/circuit_overview.md
```diff
@@ -0,0 +1,267 @@
+## Airbender Circuit Overview
+
+This document provides a high level overview of Airbender proving system for RISC‑V execution and then dives into key design patterns and stages. If you just want the big picture, read this section. Later sections explain how each part is implemented and optimized.
```
antoniolocascio, line 3: Is it an overview of the proof system or just the circuits? I think the former, so the title of the document is misleading

docs/circuit_overview.md
```diff
@@ -0,0 +1,267 @@
...
+
+State and data model
+  - Minimal explicit state per row (typically only `pc` split into 16‑bit limbs). All register/memory values flow through a global shuffle RAM argument.
+  - Instruction fetches are separate ROM lookups keyed by `pc` (ROM != RAM). Branching/jumps update `pc` deterministically. `x0` is enforced to remain zero.
+  - Execution is chunked: each chunk has “lazy init” (initial values/timestamps) and “teardown” (final values/timestamps), linked across chunks to ensure continuity.
+
+Constraints and arguments
+  - Generic constraints: arithmetic/logic, control semantics, decoder, range checks, and delegation infrastructure compiled from the CS.
+  - Hardcoded constraints (Stage 3): part that validates Stage 2 arguments (memory, lookups, delegation), boundary conditions, state linkage, ROM separation, lazy‑init sorting with the borrow trick, and per‑domain masking.
```
antoniolocascio, line 18: We're using "Stage N" before stages are introduced

docs/circuit_overview.md
```diff
@@ -0,0 +1,267 @@
...
+
+```262:267:zksync-airbender/cs/src/devices/optimization_context.rs
+witness_early_branch_if_possible(mask, placer, &evaluate_fn_inner);
+```
+
+This pattern yields significant savings:
+
+- Fewer muxes and selectors: selection happens once per relation type, not per wire.
+- Variable reuse: preallocated registers/flags/bytes are recycled across operations.
+- Betterfor GPU. Uniform relation shapes are easier to fold in Stage 3.
```
antoniolocascio, line 83:
```suggestion
- Better for GPU. Uniform relation shapes are easier to fold in Stage 3.
```

antoniolocascio (commented):
I did a pass over the docs, thanks for writing them! Other than the comments I left, I feel there's some repetition that can be avoided (machine configs, delegations). I think this would become clearer with some more structure in the docs themselves.

docs/README.md
```diff
@@ -8,4 +8,18 @@ This directory contains documentation and descriptions of the logic and design b
 - [Subarguments used](./subarguments_used.md)
```
vladbochok, line 8: `- [Custom assumptions](./custom_assumptions.md)` doesn't exist

examples/big_int/.cargo/config.toml
jcsec-security, line 1: The `big_int` example folder is missing the README. 

In addition to that, I have some comments on the other existing examples, as those are not modified in this PR I will include them here as free text:
- `basic_fibonacci`: 
    - The running and proving commands should be added for completeness and consistency, even if they are straightforward in this case
- `dynamic_fibonacci`:
    - The first command throws an err, it is missing `cli` as in `cargo run --profile cli --package cli -- run --bin [...]`
    - Same with the second one (GPU)
    - Maybe good to add the proving on CPU command at least on the first one too `cargo run --release -p cli prove --bin [...]`
    - Same with the third one (larger case)
    - The `Rebuilding` section that is located at the end of this doc should be present on the rest of the examples too as part of "the template" for completeness.
- `experiments` 
    - Inheritance errors on `Cargo.toml` when compiling. Either fix the format or hardcode the values as the rest of the examples.
    - This readme in general is a bit weak compared to the others. I think it is worth explicitly mentioning when to run `objdump.sh` and `objdump_manual.sh` for compilation
    - The first command throws and err, maybe `cargo objcopy --release -- -O binary app.bin`
    - The last paragraph references the `mini` option for machine, which doesn't seem to be supported now.
    - Worth adding the full proving command too
- `hashed_fibonacci`
    - First and second commands missing  `-p cli` 

In general, I haven't checked the GPU proving options due to hardware limitations

jcsec-security (commented):
Some comments separated from typos/format/small changes and rephrasing that I will be pushing on a different PR for @Fitznik to review beforehand

docs/tests_overview.md
```diff
@@ -0,0 +1,108 @@
...
+```bash
+cargo test -p fft   
+```
+
+---
+
+## 6. Prover integration ests (`prover_examples`)
+
+Crate **`circuit_defs/prover_examples`** contains:
+2. Runs the full staged prover pipeline.
```
jcsec-security, line 101: 3, 4 and 5 would benefit from using part of the structure of 1 and 2 to give a more clear context and so on. The "Purpose - Location - How it Works" trio, with any extra sections needed looks like a good one to follow during the whole document for consistency

docs/circuit_overview.md
```diff
@@ -0,0 +1,267 @@
...
+- Stage 4
+  - Sample a random point z in Ext4.
+  - Use barycentric evaluation to get evaluations of trace and argument polynomials at z.
+  - Construct the DEEP quotient on the main domain from these evaluations.
+  - LDE the DEEP quotient to the coset domain and commit on both domains.
+
+- Stage 5
+  - Perform FRI folding to build all FRI layers and commit to them.
+  - Draw a challenge based on these commitments, combine it with a PoW nonce, and grind to obtain a valid seed for FRI queries.
+  - Draw query points and construct queries.
```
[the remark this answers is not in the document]
jcsec-security: In general, these stages are heavy on specialized knowledge. We can either expand the glossary further or add further info here in-line as it is really the only place where some of these concepts are mentioned

PR #81, opened 2025-07-21
fix: do not use u16 -> u32 parsing for timestamps

shamatar:
## What ❔

Previously timestamps were 16 bits hardcoded, not it's configurable, but the function used legacy assumptions

## Why ❔

Correctness

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


shamatar:
@mm-zk can we move reproduction into CI job that outputs artifacts instead?

robik75 (approved):

PR #88, opened 2025-07-25
feat: allow to use generic (but recent) nightly again. Remove let chains

shamatar:
## What ❔

Mainly remove `array_chunks`

## Why ❔

Not tie to specific nightly and move along with compiler

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist


shamatar:
@mm-zk repro script is not repro enough, so it's ignored for now

PR #102, opened 2025-08-21
New recursion modes using reduced 2^23 machine

olesHolem:
## What ❔
Added new modes using reduced 2^23 machine. This makes recursion faster and cheaper. Also, it's possible to prove final recursion layers with GPU.


tools/cli/src/prover_utils.rs
```diff
@@ -35,13 +35,121 @@ pub const DEFAULT_CYCLES: usize = 32_000_000;
 #[derive(Clone, Debug, ValueEnum)]
 pub enum ProvingLimit {
     /// Does base + recursion (reduced machine).
-    FinalRecursion,
+    RecursionProof,
```
mm-zk, line 38: are you sure you want to change this name in this PR? as this means that you'll have to also update README.md and github tests.

I'd suggest doing it in separate PR after.

tools/cli/src/prover_utils.rs
```diff
@@ -35,13 +35,121 @@ pub const DEFAULT_CYCLES: usize = 32_000_000;
...
 }
 
+/// We have two layers of recursion:
+/// 1. Reduced machine (2^22 cycles) + blake delegation
+/// 2. Here we have two options:
+///   - Final reduced machine (2^25 cycles)
+///   - Reduced log23 machine (2^23 cycles) + blake delegation
+/// Then we can define four recursion strategies:
+#[derive(Clone, Copy, Debug, ValueEnum)]
+pub enum RecursionMode {
```
mm-zk, line 52: nit: I'd call it "RecursionStrategy". (you even call it like this in a comment above :)

tools/cli/src/prover_utils.rs
```diff
@@ -35,13 +35,121 @@ pub const DEFAULT_CYCLES: usize = 32_000_000;
...
+
+    pub fn finish_second_recursion_layer(
+        &self,
+        proof_metadata: &ProofMetadata,
+        proof_level: usize,
+    ) -> bool {
+        let continue_second_layer = match self {
+            RecursionMode::UseFinalMachine => proof_metadata.final_proof_count > 1,
+            RecursionMode::UseReducedLog23Machine => {
+                assert!(proof_level == 0);
```
mm-zk, line 111: add small comment why (that we expected this to do second recursion layer in one shot)

tools/cli/src/prover_utils.rs
```diff
@@ -283,7 +415,11 @@ pub fn create_proofs(
     // the production critical path
     // (tracing, witness generation, proving, recursion).
     let (mut gpu_state, mut total_proof_time) = if use_gpu {
-        (Some(GpuSharedState::new(&binary)), Some(0f64))
+        let recursion_circuit_type = MainCircuitType::ReducedRiscVMachine;
```
mm-zk, line 418: comment that the first layer of recursoin always starts with reduced machine

tools/cli/src/prover_utils.rs
```diff
@@ -368,46 +517,37 @@ pub fn load_binary_from_path(path: &String) -> Vec<u32> {
...
     const RECURSION_BINARY_KEY: usize = 1;
 
     #[cfg(feature = "gpu")]
-    pub fn new(binary: &Vec<u32>) -> Self {
-        use gpu_prover::circuit_type::MainCircuitType;
+    pub fn new(binary: &Vec<u32>, recursion_circuit_type: MainCircuitType) -> Self {
         use gpu_prover::execution::prover::ExecutableBinary;
         use gpu_prover::execution::prover::ExecutionProver;
+
+        assert!(
```
mm-zk, line 538: mention that we dont support "final" (a.k.a 25) in GPU

tools/cli/src/prover_utils.rs
```diff
@@ -679,26 +889,55 @@ pub fn create_recursion_proofs(
...
+    recursion_mode: RecursionMode,
+    use_gpu: bool,
+) -> ProgramProof {
     let (proof_metadata, proof_list) = proof_list_and_metadata_from_program_proof(input);
 
-    create_final_proofs(proof_list, proof_metadata, &None)
+    let (mut gpu_state, mut total_proof_time) = if use_gpu {
+        // Here we use GPU for final recursion layer only.
+        use gpu_prover::circuit_type::MainCircuitType;
+        let recursion_circuit_type = MainCircuitType::ReducedRiscVLog23Machine;
```
mm-zk, line 910: shoulnd't we check the recursion mode ? and if it requires "25 a.k.a. final" - then fail, saying that it is not supported ?

tools/cli/src/prover_utils.rs
```diff
@@ -334,8 +477,14 @@ pub fn create_proofs(
...
             }
             ProvingLimit::FinalProof => {
-                let program_proof =
-                    create_final_proofs(recursion_proof_list, recursion_proof_metadata, tmp_dir);
+                let program_proof = create_final_proofs(
+                    recursion_proof_list,
+                    recursion_proof_metadata,
+                    recursion_mode,
+                    tmp_dir,
+                    &mut None,
```
mm-zk, line 485: so in this case, when I run the cli until "final-proof", the last part will always be done on CPU ? (even if this is reduced23?)

but also passing the existing gpu state will be tricky, as you already probably defined it with "reduced" machine right ?

BTW - this is fine to fix in the next PR - but please add a clear comment

gpu_prover/src/prover/context.rs
```diff
@@ -47,7 +47,7 @@ impl Default for ProverContextConfig {
             powers_of_w_coarse_log_count: 12,
             allocation_block_log_size: 22,    // 4 MB blocks
             device_slack_blocks_count: 64,    // 256 MB slack
-            host_allocator_blocks_count: 128, // 512 MB host allocator pool
+            host_allocator_blocks_count: 512, // 512 MB host allocator pool
```
mm-zk, line 50: you might want to update the comment.

tools/cli/src/generate_constants.rs
```diff
@@ -1,115 +1,267 @@
...
+                _ => panic!("This recursion mode is not supported for non-universal verifier."),
+            }
         }
     };
 
     println!("End params: {:?}", end_params);
     println!("Aux values: {:?}", aux_values);
 }
 
+// pub fn generate_params_and_register_values(
```
mm-zk, line 196: please remove commented out code

tools/cli/src/generate_constants.rs
```diff
@@ -1,115 +1,267 @@
...
+//         generate_params_for_binary(first_recursion_layer_bin, Machine::Reduced);
 
-    let first_recursion_layer_end_params =
-        generate_params_for_binary(first_recursion_layer_bin, Machine::Reduced);
+//     let next_recursion_layer_end_params =
+//         generate_params_for_binary(next_recursion_layer_bin, Machine::Reduced);
 
-    let next_recursion_layer_end_params =
-        generate_params_for_binary(next_recursion_layer_bin, Machine::Reduced);
+//     let first_final_recursion_end_params =
```
mm-zk, line 244: and here

tools/cli/src/generate_constants.rs
```diff
@@ -1,115 +1,267 @@
...
+//     (end_params, aux_registers_values)
+// }
+
 pub fn generate_params_and_register_values(
-    base_layer_bin: &[u8],
-    first_recursion_layer_bin: &[u8],
-    next_recursion_layer_bin: &[u8],
-    first_final_recursion_bin: &[u8],
-    next_final_recursion_bin: &[u8],
+    machines_chain: &[(&[u8], Machine)],
```
mm-zk, line 218: if you're simplifying it already -- why not just get machines_chain ?

(and inside you can extract that last_machine = machines_chain[-1]

circuit_defs/setups/src/lib.rs
```diff
@@ -57,47 +58,47 @@ pub fn is_final_reduced_machine_configuration<C: MachineConfig>() -> bool {
...
-        reduced_risc_v_machine::LDE_FACTOR
-    } else if is_final_reduced_machine_configuration::<C>() {
-        final_reduced_risc_v_machine::LDE_FACTOR
-    } else if is_machine_without_signed_mul_div_configuration::<C>() {
-        machine_without_signed_mul_div::LDE_FACTOR
-    } else {
-        panic!("unknown machine configuration {:?}", C::default())
-    }
-}
+// pub fn num_cycles_for_machine<C: MachineConfig>() -> usize {
```
mm-zk, line 61: remove commented out code

circuit_defs/prover_examples/src/lib.rs
```diff
@@ -171,7 +172,8 @@ pub fn trace_execution_for_gpu<
     HashMap<u16, Vec<DelegationWitness>>,
     Vec<FinalRegisterValue>,
 ) {
-    let cycles_per_circuit = setups::num_cycles_for_machine::<C>();
+    // let cycles_per_circuit = setups::num_cycles_for_machine::<C>();
```
mm-zk, line 175: cleanup

circuit_defs/prover_examples/src/lib.rs
```diff
@@ -467,7 +481,7 @@ pub fn prove_image_execution_for_machine_with_gpu_tracers<
             },
         };
 
-        let lde_factor = setups::lde_factor_for_machine::<C>();
+        // let lde_factor = setups::lde_factor_for_machine::<C>();
```
mm-zk, line 484: cleanup

circuit_defs/trace_and_split/src/lib.rs
```diff
@@ -19,7 +19,7 @@ use setups::prover::risc_v_simulator::abstractions::non_determinism::*;
 use setups::prover::risc_v_simulator::cycle::MachineConfig;
 use setups::prover::transcript::Seed;
 use setups::prover::*;
-use setups::trace_len_for_machine;
+// use setups::trace_len_for_machine;
```
mm-zk, line 22: cleanup

circuit_defs/trace_and_split/src/lib.rs
```diff
@@ -319,7 +319,8 @@ pub fn commit_memory_tree_for_riscv_circuit_using_gpu_tracer<C: MachineConfig, A
     lde_precomputations: &LdePrecomputations<A>,
     worker: &Worker,
 ) -> (Vec<MerkleTreeCapVarLength>, WitnessEvaluationAuxData) {
-    let lde_factor = setups::lde_factor_for_machine::<C>();
+    // let lde_factor = setups::lde_factor_for_machine::<C>();
```
mm-zk, line 322: and here

circuit_defs/trace_and_split/src/lib.rs
```diff
@@ -586,7 +588,7 @@ pub fn run_and_split_for_gpu<
         setups::risc_v_cycles::ROM_ADDRESS_SPACE_SECOND_WORD_BITS,
         setups::final_reduced_risc_v_machine::ROM_ADDRESS_SPACE_SECOND_WORD_BITS
     );
-    let domain_size = trace_len_for_machine::<C>();
+    // let domain_size = trace_len_for_machine::<C>();
```
mm-zk, line 591: and here

gpu_prover/src/prover/context.rs
olesHolem: Actually, I think it should remain 128. I will test it one more time.

tools/cli/src/generate_constants.rs
olesHolem: I prefer to have them separately because they play different roles. For example, for UseReducedLog23Machine strategy, the chain will look like: '[(base_layer_bin, Machine::Standard), (execution_utils::UNIVERSAL_CIRCUIT_VERIFIER, Machine::Reduced), (execution_utils::UNIVERSAL_CIRCUIT_VERIFIER, Machine::ReducedLog23)]' and for UseReducedLog23MachineMultiple it will be: '[(base_layer_bin, Machine::Standard), (execution_utils::UNIVERSAL_CIRCUIT_VERIFIER, Machine::Reduced), (execution_utils::UNIVERSAL_CIRCUIT_VERIFIER, Machine::ReducedLog23), (execution_utils::UNIVERSAL_CIRCUIT_VERIFIER, Machine::ReducedLog23)]'.
So we would need to duplicate the same (binary, machine) pair in function parameter, where we avoid duplications. IMHO, it's more intuitive to keep them separately. The key point here is that last machine does not count during computation of `end_params`, so that's why we track if there is exactly one repetition of last machine or at least to.

mm-zk: Good point. Please mention it in the comment to this function.

tools/cli/src/prover_utils.rs
mm-zk: thanks for adding the comment

PR #106, opened 2025-08-27
chore: moving functions from cli_utils to execution_utils

mm-zk:
## What ❔

* Moving some methods from cli utils to execution utils, as they are used outside of airbender too.
* No logic change - just code moving.

## Why ❔

* CLI util was originally built 'just' for the cli - but as the functionality evolved, more external projects (zkos server, zkos wrapper) started depending on this crate.
* instead, this logic should be inside the 'execution_utils' crate (which is has a better name and location).


execution_utils/src/structs.rs
```diff
@@ -0,0 +1,284 @@
+use std::{collections::BTreeMap, path::Path};
+
+use trace_and_split::FinalRegisterValue;
+use verifier_common::{
+    blake2s_u32::BLAKE2S_DIGEST_SIZE_U32_WORDS, cs::utils::split_timestamp,
+    prover::prover_stages::Proof,
+};
+
+// Structures to serialize / deserialize airbender proofs.
+// ProgramProof = ProofMetadata + ProofList.
```
0xVolosnikov, line 10: Can we use same names for fields then?

Also unclear then, why we do have `reduced_proofs` in `ProofList`, but don't have in `ProgramProof`

execution_utils/src/structs.rs
```diff
@@ -0,0 +1,284 @@
+use std::{collections::BTreeMap, path::Path};
```
0xVolosnikov, line 1: Looks like this file is fully dedicated to Proofs. Maybe `structs` name is too generic in this case?

execution_utils/src/structs.rs
mm-zk: I know that there are a bunch of improvements needed - but I purposefully split it into two PRs.

This one is just moving things around, without any logic change, and the follow-up one will change the logic.

execution_utils/src/structs.rs
```diff
@@ -0,0 +1,284 @@
...
+            end_params: proof_metadata.end_params,
+            recursion_chain_preimage: proof_metadata.prev_end_params_output,
+            recursion_chain_hash: proof_metadata.prev_end_params_output_hash,
+        }
+    }
+    pub fn to_metadata_and_proof_list(self) -> (ProofMetadata, ProofList) {
+        let reduced_proof_count = self.base_layer_proofs.len();
+        let proof_list = ProofList {
+            basic_proofs: vec![],
+            // Here we're guessing - as ProgramProof doesn't distinguish between basic and reduced proofs.
```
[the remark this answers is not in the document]
mm-zk: I know that there are a bunch of improvements needed - but I purposefully split it into two PRs.

This one is just moving things around, without any logic change, and the follow-up one will change the logic.

execution_utils/src/verifiers.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    assert_eq!(32, metadata.register_values.len());
+    for register in metadata.register_values.iter() {
+        oracle_data.push(register.value);
+        let (low, high) = split_timestamp(register.last_access_timestamp);
+        oracle_data.push(low);
+        oracle_data.push(high);
+    }
+
+    let delegations: Vec<u32> = if metadata.basic_proof_count > 0 {
```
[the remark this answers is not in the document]
mm-zk: yes, that's the weakness of ProgramProof -- will be fixed in the next PR.

execution_utils/src/structs.rs
mm-zk: good point - I'll rename to proofs.

RomanBrodetski (approved):

PR #112, opened 2025-09-04
feat: move constants into separate crate for external imports

shamatar:


shamatar:
Now also contains minimal part of #111

PR #132, opened 2025-09-24
feat!: removing final machine support

mm-zk:
## What ❔

* Removing support for 2^25 'final' machine
* We use 2^23 with delegations instead.


execution_utils/src/proofs.rs
```diff
@@ -30,7 +30,11 @@ pub struct ProofMetadata {
     pub basic_proof_count: usize,
     pub reduced_proof_count: usize,
```
shamatar, line 31: Not needed anymore?

shamatar (approved):

PR #169, opened 2025-11-18
feat: added functionality for 100 security bits

olesHolem:
## What ❔

Added functionality for controlling security bits using PoW right before drawing new challenge. Added functions to `prover/src/prover_stages/pow_bits_calculator.rs` to determine number PoW bits for each challenge in prover. Prover computes these parameters from its inputs (circuit, lde_factor, etc.) and verifier from its constant parameters. Also, we can compute "worst case scenario" parameters and reuse them for all circuits.


circuit_defs/bigint_with_control/verifier/src/lib.rs
```diff
@@ -256,29 +326,61 @@ pub unsafe fn verify_with_configuration<I: NonDeterminismSource, V: LeafInclusio
     let mut fri_folding_challenges: [Mersenne31Quartic; NUM_FRI_STEPS] =
         [MaybeUninit::uninit().assume_init(); NUM_FRI_STEPS];
 
-    for (caps, challenge) in skeleton
+    for (((caps, challenge), pow_challenge), pow_bits) in skeleton
```
shamatar: As we have too many zip-s here, and we know that iterator lengths should be same size, it's better to change it into index-based iteration

verifier_common/src/lib.rs
```diff
@@ -12,20 +12,25 @@ compile_error!("multiple security levels selected same time");
...
+);
 
 #[cfg(feature = "security_100")]
 pub const SECURITY_BITS: usize = 100;
 #[cfg(feature = "security_100")]
 pub const POW_BITS: usize = 28;
+#[cfg(feature = "security_100")]
+pub const MEMORY_DELEGATION_POW_BITS: usize = pow_bits_for_memory_and_delegation(
+    SECURITY_BITS,
+    cs::one_row_compiler::MAX_NUMBER_OF_CYCLES.leading_zeros() as usize,
```
shamatar: Does it build in no-std? If we import `cs` as "definitions only", then "one row compiler" module is not available? Also, leading zeroes are very misleading. If the number of cycles is not a power of two, instead it's better to take "next power of two", and then trailing zeroes

shamatar (changes requested):
Minor structural changes and few better comments

shamatar (approved):

PR #196, opened 2026-02-06
Sum-check loop

yoaveshel:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/prover/src/gkr/prover/mod.rs
+++ b/prover/src/gkr/prover/mod.rs
@@ -1,7 +1,7 @@
 use std::alloc::Global;
 use std::collections::BTreeMap;
 
-use cs::gkr_compiler::GKRCircuitArtifact;
+use cs::gkr_compiler::{GKRCircuitArtifact, OutputType};
 use field::TwoAdicField;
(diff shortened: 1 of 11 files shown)


prover/src/gkr/prover/mod.rs
```diff
@@ -263,33 +263,72 @@ where
...
+        claim_readset,
+        claim_writeset,
+        claim_rangechecknum,
+        claim_rangecheckden,
+        claim_timechecknum,
+        claim_timecheckden,
+        claim_lookupnum,
+        claim_lookupden,
+        z
+    ) = todo!();
```
shamatar, line 310: In general you can just make them by yourself (draw random point and evaluate - you have polys in storage), so you can test your code without depending on Mario's result yet

shamatar: And we will also need small logic on how to "decompose" cache relations at base layer - basically check what missing evaluations we have, and compute them explicitly to send to the verifier (instead of verifier deriving them as a product of sumcheck)

shamatar:
I'll accept it and will add a stub myself

PR #197, opened 2026-02-09
feat: Only use inlining in verifiers when compile for RISC-V

popzxc:
Extra changes:
- Fixed reproducible build
- Pinned nightly version for verifier.

On native, we don't really care about performance of the verifier, but it's important when compiled to RISC-V.
This PR makes is so that on native we never inline field ops, but always do on RISC-V.

With this change, I am able to build a project which depends on both prover and `full_statement_verifier` in 59 seconds; previously it was several minutes even with `opt-level = 0` on verifier crates (and tens of minutes if I try to just compile to release), so the impact is very noticeable.

Verifier check via reproducible build script: 

On my branch: 

```
Successfully copied 1.25MB to /<..>/airbender/zksync-airbender/tools/verifier/
db1ffd825fe07e8b9472653b79a33fe7  tools/verifier/recursion_in_unified_layer.bin
Successfully copied 2.29MB to /<..>/airbender/zksync-airbender/tools/verifier/
b67366a47d6a51676d59dec074ae536e  tools/verifier/recursion_in_unrolled_layer.bin
```

On dev:

```
Successfully copied 1.25MB to /<..>/airbender/zksync-airbender/tools/verifier/
db1ffd825fe07e8b9472653b79a33fe7  tools/verifier/recursion_in_unified_layer.bin
Successfully copied 2.29MB to /<..>/airbender/zksync-airbender/tools/verifier/
b67366a47d6a51676d59dec074ae536e  tools/verifier/recursion_in_unrolled_layer.bin
```

I did not commit the recompiled verifiers.


tools/verifier/build.sh
```diff
@@ -28,22 +28,22 @@ rm recursion_in_unified_layer.elf
 rm recursion_in_unified_layer.text
 
 # Build something simple to check for errors
-# cargo build --release -Z build-std=core,panic_abort,alloc -Z build-std-features=panic_immediate_abort  --features unrolled_base_layer,security_80 --no-default-features
+# cargo build --release -Z build-std=core,panic_abort,alloc --features unrolled_base_layer,security_80 --no-default-features
```
popzxc: There is already `panic=abort` in the `Cargo.toml`, and `panic_immediate_abort`. I believe in the old version of `cargo objcopy` this flag was simply ignored, but now it triggers a compile error.

tools/reproduce/reproduce.sh
```diff
@@ -12,26 +12,28 @@ docker build -t airbender-verifiers  -f tools/reproduce/Dockerfile .
...
-    universal.reduced.vk.json
-    universal_no_delegation.final.vk.json
-    recursion_layer.reduced.vk.json
-    recursion_layer_no_delegation.final.vk.json
-    final_recursion_layer.final.vk.json
-    universal.reduced_log23.vk.json
-    recursion_log_23_layer.reduced.vk.json
+    recursion_in_unified_layer.bin
+    recursion_in_unrolled_layer.bin
+    # base_layer.bin
```
popzxc: Not sure if I should leave these commented out (for the future? don't know) if to remove them.

tools/verifier/rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly"
\ No newline at end of file
+channel = "nightly-2026-01-19"
```
popzxc: Reproducible build mentions that the version should be pinned & equal to this one, so I guess it must be pinned.
So I pinned to the one I use locally now.

tools/verifier/build.sh
```diff
@@ -71,36 +71,36 @@ rm recursion_in_unified_layer.text
...
 
 #cargo build -Z build-std=core,panic_abort,alloc -Z build-std-features=panic_immediate_abort --release --no-default-features --features=recursion_in_unified_layer,security_80
-(CARGO_TARGET_DIR=target/seventeen cargo objcopy --release -Z build-std=core,panic_abort,alloc -Z build-std-features=panic_immediate_abort  --features recursion_in_unified_layer,security_80 --no-default-features -- -O binary recursion_in_unified_layer.bin;
-CARGO_TARGET_DIR=target/seventeen cargo objcopy --release -Z build-std=core,panic_abort,alloc -Z build-std-features=panic_immediate_abort  --features recursion_in_unified_layer,security_80 --no-default-features -- -R .text recursion_in_unified_layer.elf;
-CARGO_TARGET_DIR=target/seventeen cargo objcopy --release -Z build-std=core,panic_abort,alloc -Z build-std-features=panic_immediate_abort  --features recursion_in_unified_layer,security_80 --no-default-features -- -O binary --only-section=.text recursion_in_unified_layer.text) &
+(CARGO_TARGET_DIR=target/seventeen cargo objcopy --release -Z build-std=core,panic_abort,alloc --features recursion_in_unified_layer,security_80 --no-default-features -- -O binary recursion_in_unified_layer.bin;
```
shamatar: Can you add another two builds to with `security_100` feature, and name binaries like `recursion_in_unrolled_layer_100_bits.*` ?

rust-toolchain.toml
```diff
@@ -1,3 +1,2 @@
 [toolchain]
-channel = "nightly"
-
+channel = "nightly-2026-01-19"
```
shamatar: The outer one I'd prefer not to pin in general (we can quickly fix assert_matches imports, the rest should work)

shamatar: verifier crate has different toolchain pinning to manage, but this one is for workspace in full

tools/reproduce/reproduce.sh
shamatar: you can remove `base_layer` and similar. In general we will need to delete old circuit definitions, and so matching recursive verifiers

shamatar (approved):

PR #204, opened 2026-02-16
Add self-checks, integrate transcript into the sumcheck loop, debug kernels

yoaveshel:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


prover/src/gkr/prover/sumcheck_loop/mod.rs
```diff
@@ -0,0 +1,368 @@
...
+    let transcript_inputs: Vec<E> = last_evaluations
+        .iter()
+        .map(|el| el.1.iter())
+        .flatten()
+        .copied()
+        .collect();
+    commit_field_els(seed, &transcript_inputs);
+
+    let challenges = draw_random_field_els::<F, E>(seed, 2);
+    let [r_before_last, r_last, next_batching_challenge] = challenges.try_into().unwrap();
```
yoaveshel, line 89: `gkr_run_basic_unrolled_test` fails because we draw `r_before_last` and `r_last` twice - once here and once in `run_sumcheck_loop`, so `folding_challenges` is too long

shamatar: I had it fixed (draw less in sumcheck loop), as we can only draw last challenge after presenting last evaluations (explicit form). But I'll double-check

shamatar: Actually you just merged it wrong - I had it all implemented

PR #205, opened 2026-02-16
fix: regenerate source files and binaries

robik75:
## What ❔

Regenerates checked-in generated artifacts and verifier binaries:
- circuit `generated/quotient.rs` files (and verifier counterparts),
- `cs` layout/SSA JSON files,
- verifier binaries (`.bin/.elf/.text`).

## Why ❔

To keep generated outputs in sync with the current toolchain/generators and restore a consistent repo state for CI and downstream users.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


robik75:
@shamatar @popzxc heads up, the github-actions bot has rewritten the verifier binaries which broke the verification, i had to revert that, the bot needs fixing

robik75:
i had to disable the  update-verifiers workflow

popzxc (approved):
I'm OK with it as an interim solution, but we need to fix the workflow -- if reproducible builds don't work as expected, we need to fix it ASAP.

PR #210, opened 2026-02-23
feat[gpu_prover]: try acting on 10 columns (depth-first) instead of 1 in hypercube bench

mcarilli:
## What ❔

"Augments" hypercube bench to act on 10 columns depth-first, instead of 1 column repeatedly.

## Why ❔

Possibly more realistic, and eliminates spurious L2 residency of a single column across bench iterations.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/gpu_prover/benches/hypercube.rs
+++ b/gpu_prover/benches/hypercube.rs
@@ -15,6 +15,7 @@ use gpu_prover::{
 };
 
 const LOG_ROWS: u32 = 24;
+const COLS: usize = 10;
 
 struct HypercubeBitrevBenchCase {
     rows: usize,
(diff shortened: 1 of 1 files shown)


gpu_prover/benches/hypercube.rs
```diff
@@ -26,28 +27,34 @@ impl HypercubeBitrevBenchCase {
     fn new(stream: &CudaStream) -> CudaResult<Self> {
         let rows = 1usize << LOG_ROWS;
 
-        let mut d_src = DeviceAllocation::alloc(rows)?;
-        let d_dst = DeviceAllocation::alloc(rows)?;
+        let mut d_src = DeviceAllocation::alloc(rows * COLS)?;
+        let d_dst = DeviceAllocation::alloc(rows * COLS)?;
 
         // Fill once to avoid benchmarking uninitialized memory reads.
```
shamatar, line 33: Believe me, you can get into benchmarking allocator here - as it's zeroed memory and it may be allocated lazily

mcarilli: It's a good point that i should ensure all first touch overhead happens before benchmarking iterations, but this particular spot allocates gpu memory which doesn't have the same lazy first touch behavior as cpu ram.

Anyway, RobertGPT fixed the whole issue a different way [here](https://github.com/matter-labs/zksync-airbender/commit/7ab528b0159fe8643d484ee1785f4f8d0cf791e9), so my patch is no longer needed.

mcarilli:
Fixed by Robert in https://github.com/matter-labs/zksync-airbender/commit/7ab528b0159fe8643d484ee1785f4f8d0cf791e9

PR #218, opened 2026-02-26
perf: add more parallelisation to gkr prover

yoaveshel:
## What ❔
Reduces proving time by 5s (from 48s to 43s) compared to latest av_gkr_compiler, when running `gkr_run_basic_unrolled_test` without self-checks

- Adds `distribute_powers_parallel` to fft (not used)
- Parallelise per coset LDE computation for whir                                                                        
- Precomputes combined_challenges = r1 * r2 in BaseFieldPolySourceAfterTwoFoldings to eliminate a redundant extension field multiplication on every get_at_index call during the two-folding step

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


shamatar:
Yoav, we have two different cases for LDE computations:
- during commit we assume multiple polynomials, so each thread takes a poly, and then computes all LDEs for it. In this case we just need serial power distribution
- during WHIR we have 1 poly, and different threads should compute their own coset-FFT, so again - parallel distribution makes little sense

So are you sure that parallel distribution is needed and gives anything?

Also, would you do benchmarks without GKR self-checks feature? Self-checks are heavy, so you get very skewed estimates

yoaveshel:
@shamatar The measurements above are without self-checks (with self-checks, it takes about 100s)

If I understand correctly, then I agree that `distribute_powers_parallel` is redundent:                                                                                                                                                              
  - Commit: `lde_multiple_polys_parallel_from_hypercubes` already parallelises across polynomials, so each thread is busy with its own poly and calls `compute_column_major_lde_from_monomial_form` serially — adding intra-op parallelism just contends with the outer parallelism.
  - WHIR: the coset loop in `compute_column_major_lde_from_monomial_form` is sequential — but the right fix would be to parallelise across cosets there (each thread handles its own coset-FFT serially), not to parallelise within a single coset's `distribute_powers`.

Also, I now see that I parallelised dead code: `compute_column_major_lde_from_main_domain_inner` and `_and_output_monomial_form`, are not called anywhere. Do we plan to use them at some point, or are they leftovers? The function that actually runs is `compute_column_major_lde_from_monomial_form`, which is unchanged.

Should I add coset-level parallelisation for the WHIR case?
And if we do plan to use `compute_column_major_lde_from_main_domain_inner` and `_and_output_monomial_form` later, should I write a `distribute_powers_bitreversed` that operates in bitreverse order? Or did you have another optimisation in mind?

shamatar:
They are "debugging" leftovers. If you implemented some speed for them - it's fine, but I do not think we will need them in workflow

shamatar:
Yoav, will you update this PR, or split/make another one?

shamatar (approved):

PR #220, opened 2026-02-26
feat(cli,zksmith)!: migrate prover flow, add security_100, re-enable CI

robik75:
## What ❔

- Reworked tools/cli around the current proving API (`ProgramProver` + artifact-based flow) and removed legacy CLI paths/files (`setup.rs`, `vk.rs`).
- Added explicit security-level feature wiring across the stack:
  - `security_80` (default)
  - `security_100`
  - matching verifier feature variants (`include_verifiers_80`, `include_verifiers_100`).
- Updated recursion-unified behavior in CLI/GPU orchestration:
  - `RecursionUnified` verification uses unified-path semantics.
  - unified recursion convergence is driven by proof-shape target count.
- Re-enabled and stabilized CI for the new flow:
  - dedicated transpiler test job with an allowlist,
  - security_100 recursion-unified smoke coverage,
  - GPU workflow hardening (matrix/runtime tuning and multi-arch binary handling),
  - verifier-related workflows aligned to current runner usage.
- Updated CLI and zksmith docs to reflect the current commands/feature model.

## Why ❔

`dev` had diverged from the current prover architecture, so CLI/zksmith behavior and related CI were no longer aligned with the codebase. This PR brings tooling, feature wiring, recursion handling, and CI back in sync with the current implementation.

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist


popzxc (approved):
Generally LGTM. I'm not sure what would be the right way to proceed with CLI now that we have cargo airbender, it feels somewhat redundant to maintain two things; but at the same time cargo airbender is in a separate repo. I'll think about how to organize it better in the future, but for now it's certainly an improvement.

yoaveshel (approved):

PR #222, opened 2026-03-09
test: Migrate opcode tests to the transpiler crate

popzxc:
1. Removes old `opcode_tests` crate -- it was disabled and broken already, and it will likely be broken by the new proving system anyway; so it'll be easier to reimplement it from scratch when new proving system is ready.
2. Migrates a part of the old tests to the transpiler crate -- for things that do not have coverage yet and where it makes sense.


shamatar:
Do you want to also add it to CI?

popzxc:
@shamatar it's [already in CI](https://github.com/matter-labs/zksync-airbender/actions/runs/22851922607/job/66282325972?pr=222#step:6:1118) (output is noisy because we're running with `--nocapture` and in rejection tests we're catching panics -- but everything passes)

shamatar (approved):

PR #231, opened 2026-03-12
chore: Rework CI (dev branch)

popzxc:
This PR:
- Does similar refactoring to one in #229 -- jobs are now consolidated under `pr-checks` / `gpu-checks`.
- Removes `docker/anvil-bender` & workflow
- Removed `zksmith` -- now for GPU tests we use CLI and run the same program (hashed-fibonacci)
  - Bonus: now in GPU tests we actually verify generated proof as an extra smoke test
- Reworked CLI to support continuing proving from the previous stage. We used to have 2 separate jobs to calculate & verify base proof, and same for unified recursion. But unified recursion _already_ contains base layer, so with that we got rid of 2 jobs.
- Removed `build_verifier` job -- with `ckech_verifiers` it's meaningless.
- Verifiers are now always included in CLI -- since compile time for them improved, it doesn't make much sense to have a separate mode without verifiers. This is also +1 removed job from CI.
- Uses faster runner for building verifier for GPU test


tools/cli/Cargo.toml
```diff
@@ -50,45 +50,36 @@ default = ["security_80"]
 
 security_80 = [
     "execution_utils/security_80",
+    "execution_utils/verifier_80",
```
robik75, line 53: i believe `execution_utils/verifier_80` already includes `execution_utils/security_80`

tools/cli/Cargo.toml
```diff
@@ -50,45 +50,36 @@ default = ["security_80"]
...
-    "execution_utils/verifier_80",
+    "gpu_prover?/security_80",
+    "full_statement_verifier?/security_80",
+    "blake2_with_compression_verifier?/security_80",
 ]
-include_verifiers_100 = [
-    "security_100",
+security_100 = [
+    "execution_utils/security_100",
+    "execution_utils/verifier_100",
```
robik75, line 69: same here

tools/cli/Cargo.toml
popzxc: I think it's fine being explicit here, just in case. The feature layout in the code is pretty convoluted, and it brings no harm.

robik75 (approved):

PR #232, opened 2026-03-13
chore: Remove more legacy tests

popzxc:
Removes old simulator-based tests for GPU.

Instead, adds more or less equivalent ones that are based on the transpiler.

I have manually verified that newly-added tests pass.


gpu_prover/src/prover/stage_2_kernels.rs
```diff
@@ -1013,27 +1011,4 @@ pub(crate) mod tests {
...
-    #[serial]
-    #[ignore]
-    fn test_standalone_stage_2_non_unrolled_for_main_and_keccak() {
-        let ctx = DeviceContext::create(12).unwrap();
-        run_keccak_test_impl(
-            Some(Box::new(comparison_hook)),
-            Some(Box::new(comparison_hook)),
-        );
-        ctx.destroy().unwrap();
-    }
```
popzxc, line 1038: Note that we now use the hook in `unrolled_prover/stage_2_kernels.rs`.

robik75 (dismissed):

gpu_prover/src/prover/stage_2_kernels.rs
robik75: ACK

robik75 (dismissed):

robik75 (approved):

PR #234, opened 2026-03-16
test: Add corner case tests for bigint delegation in unrolled circuit

Fitznik:
## What ❔

Added 196 corner case tests for the bigint_with_control delegation circuit in the unrolled proving pipeline. 

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


examples/bigint_with_control/src/main.rs
```diff
@@ -16,41 +17,1446 @@ unsafe extern "C" fn start_rust() -> ! {
     main()
 }
 
-#[repr(C, align(32))]
-struct AlignedU256 {
-    words: [u32; 8],
+#[inline(always)]
```
popzxc: Why do we need to copy-paste it from common constants? IMHO less duplication is better.

examples/bigint_with_control/src/main.rs
```diff
@@ -16,41 +17,1446 @@ unsafe extern "C" fn start_rust() -> ! {
...
-    zksync_os_finish_success(&[
-        accumulator.words[0],
-        accumulator.words[1],
-        accumulator.words[2],
-        carry,
-        0,
-        0,
-        0,
-        0,
+    fn run_case(a: &mut U256, b: &U256, out: &mut [u32], mut mask: u32) {
```
popzxc: Can we create a module e.g. `test_suite.rs`, encapsulate each test case into a separate function with doc-comment, and provide a method e.g. `test_suite::run_edge_case_tests()` which will be called here? It'll be much more clear.

Also, I believe that each test case can be expressed as a set of inputs in a test vector, e.g.

```rust
struct TestInput { .. }
type ExpectedOutput = U256;

const TEST_VECTORS: [(TestInput, ExpecteOutput); ..] = ...;

fn run_edge_case_tests() {
  for (input, expected_output) in TEST_VECTORS { .. }
}
```

examples/bigint_with_control/Cargo.toml
```diff
@@ -10,7 +10,7 @@ keywords = ["blockchain", "zksync", "zk", "risc-v"]
 categories = ["cryptography"]
 
 [dependencies]
-common_constants = { path = "../../common_constants" }
+common_constants = { path = "../../common_constants"}
```
popzxc, line 13: Nit: I'd suggest undoing this change -- space between `}` is idiomatic

examples/bigint_with_control/src/test_suite.rs
```diff
@@ -0,0 +1,1508 @@
...
+///
+/// Operation masks are built from bit indices defined in `common_constants`:
+///   - bit 0: ADD
+///   - bit 1: SUB
+///   - bit 2: SUB_AND_NEGATE (computes b - a)
+///   - bit 3: MUL_LOW
+///   - bit 4: MUL_HIGH
+///   - bit 5: EQ
+///   - bit 6: CARRY / BORROW-IN
+///   - bit 7: MEMCOPY
```
popzxc: Nit: this should be `//!`, not `///` -- `///` only adds doc-comment for the item below (e.g. to this import), while `//!` adds comment to the outer item (e.g. module).

popzxc (commented):
Mostly LGTM, one small nit 😅

popzxc (approved):
Thanks!

shamatar (approved):

PR #237, opened 2026-03-17
feat: Implement support for cycle markers in transpiler

popzxc:
This PR implements cycle marker API that previously was available for the simulator.

For that, we extend VM with one more generic parameter -- `ExecutionObserver` -- which is meant to do any general-purpose (not proving-related) observations. It is done in a non-dyn-friendly way, but this way we have efficiently 0 footprint for the proving path.

The API can be used as follows:

```rust
let (_finished, cycle_markers) = CycleMarkerHooks::with(|| {
    VM::<DelegationsCounters, CycleMarkerHooks>::run_basic_unrolled::<_, _, _>(
        &mut state,
        &mut ram,
        &mut (),
        &tape,
        program.len(),
        &mut (),
    )
});
```

### DIFF
--- a/common_constants/src/lib.rs
+++ b/common_constants/src/lib.rs
@@ -5,6 +5,17 @@ pub mod delegation_types;
 pub mod rom;
 pub mod timestamps;
 
+pub mod dev {
+    /// Development-only CSR recognized by the transpiler-side cycle marker hooks.
+    ///
+    /// This is intended for local transpiler profiling only.
+    ///
+    /// The proving path rejects this CSR during replay/witness generation, so a
+    /// program that contains it should be treated as a development artifact and
+    /// must not be proved.
+    pub const TRANSPILER_MARKER_CSR: u32 = 0x7ff;
+}
+
 pub use self::circuit_families::*;
 pub use self::delegation_types::*;
 pub use self::rom::*;
--- a/common_constants/src/lib.rs
+++ b/common_constants/src/lib.rs
@@ -5,7 +5,9 @@ pub mod delegation_types;
 pub mod rom;
 pub mod timestamps;
 
-pub mod dev {
+/// This module is meant to contain extensions that are outside of the proving path,
+/// e.g. for development.
+pub mod internal_features {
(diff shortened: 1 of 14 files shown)


common_constants/src/lib.rs
```diff
@@ -5,6 +5,17 @@ pub mod delegation_types;
 pub mod rom;
 pub mod timestamps;
 
+pub mod dev {
```
shamatar: just name is `marker_csr` or `internal_features` - it's going to stay and doesn't deserve just `dev`

riscv_transpiler/src/ir/mod.rs
```diff
@@ -648,6 +648,18 @@ pub fn preprocess_bytecode<OPT: DecodingOptions>(bytecode: &[u32]) -> Vec<Instru
                             // short-cut
                             continue;
                         }
+                        common_constants::dev::TRANSPILER_MARKER_CSR => {
```
shamatar: I think we should assert rs1 == 0 and rd == 0 here, so canonical opcode is `csrrw x0, 0x7ff, x0`

antoniolocascio (approved):
LGTM, thanks!

riscv_transpiler/src/cycle/markers.rs
```diff
@@ -0,0 +1,169 @@
...
+
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct Mark {
+    pub cycles: u64,
+    pub delegations: HashMap<u32, u64>,
+}
+
+impl Mark {
+    pub fn diff(&self, before: &Self) -> Self {
+        let cycles = self.cycles - before.cycles;
```
shamatar, line 64: I think it's prone to off-by-one error: if we have a sequence of marker - opcode - marker, then we should return 1. It's not that important for stats as usually we profile larger functions, so you may fix or ignore

shamatar (approved):

riscv_transpiler/src/cycle/markers.rs
popzxc: Good catch, will fix in a follow-up PR.

PR #240, opened 2026-03-18
fix: reset cycle markers for fresh simulator states

ia-agentic:
## Summary
- reset the thread-local cycle marker collector whenever a fresh simulator state is created
- apply the reset in both the baseline simulator and the unrolled prover state so marker counts do not leak across runs
- add a regression test covering the one-cycle carry-over case for both simulator implementations

## Why
Cycle marker state lives in thread-local storage. Creating a fresh simulator state did not clear that storage, so a previous run could shift the next run's first marker by one or more cycles. Resetting the collector on fresh state creation makes marker counts scoped to the current execution.

## Testing
- cargo test -p risc_v_simulator
- cargo test -p risc_v_simulator --features cycle_marker

### DIFF
--- a/risc_v_simulator/src/cycle/state.rs
+++ b/risc_v_simulator/src/cycle/state.rs
@@ -222,6 +222,11 @@ thread_local! {
 }
 
 #[cfg(feature = "cycle_marker")]
+pub(crate) fn reset_cycle_marker() {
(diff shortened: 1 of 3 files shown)


popzxc:
Agent went a bit mad, sorry, ignore this

PR #250, opened 2026-03-30
fix: add local non-x86_64 gpu prover JIT stubs

ia-agentic:
## Summary
- keep `riscv_transpiler::jit` x86_64-only and remove the non-x86_64 JIT stub exports
- make `gpu_prover` use the JIT simulator on `x86_64` and fall back to the interpreter-backed simulator on non-`x86_64`
- preserve the existing trace/snapshot production flow so `ZKSYNC_USE_CUDA_STUBS` builds can still compile on unsupported architectures without introducing panicking JIT placeholders

## Problem
`gpu_prover` imported `riscv_transpiler::jit::{Context, JittedCode, ...}` unconditionally. That broke compilation on Apple Silicon before the CUDA stubs could do their job, even though the real issue is that the JIT is x86_64-only.

## Verification
- `ZKSYNC_USE_CUDA_STUBS=1 cargo check -p gpu_prover`
- `ZKSYNC_USE_CUDA_STUBS=1 cargo check -p gpu_prover --target aarch64-unknown-linux-gnu`
- attempted `cargo check -p gpu_prover --target aarch64-apple-darwin`, but this Linux runner does not have the Darwin C toolchain / SDK required by `capstone-sys` and `nvtx`


riscv_transpiler/src/jit/mod.rs
```diff
@@ -19,9 +19,21 @@ pub use self::structs::*;
 
 #[cfg(all(target_arch = "x86_64", feature = "jit"))]
 mod impls;
+#[cfg(all(
+    target_pointer_width = "64",
+    not(target_arch = "x86_64"),
+    feature = "jit"
+))]
+mod unsupported;
```
popzxc: Wouldn't it be better to change `gpu_prover` so that it uses JIT on x86_64 and falls back to non-jitted version on non-x86_64 architectures? 

If not, respond to this comment explaining why.

ia-agentic: Changed it in that direction. `gpu_prover` now keeps the JIT path on `x86_64` and falls back to the interpreter-backed simulation path on non-`x86_64`, so `riscv_transpiler` no longer exports unsupported JIT stubs for other architectures. I also checked the stubbed build with `ZKSYNC_USE_CUDA_STUBS=1 cargo check -p gpu_prover` and the non-`x86_64` path with `ZKSYNC_USE_CUDA_STUBS=1 cargo check -p gpu_prover --target aarch64-unknown-linux-gnu`.

gpu_prover/src/execution/simulation_runner.rs
```diff
@@ -108,11 +120,304 @@ pub(crate) struct Snapshot<R: DataTraceRanges> {
 
 unsafe impl<R: DataTraceRanges> Send for Snapshot<R> {}
 
+#[cfg_attr(target_arch = "x86_64", allow(dead_code))]
+pub(crate) trait VmCounters {
+    fn into_machine_counters(self) -> [u64; MAX_NUM_COUNTERS];
+}
+
+impl VmCounters for DelegationsAndFamiliesCounters {
+    fn into_machine_counters(self) -> [u64; MAX_NUM_COUNTERS] {
```
popzxc: OK, it looks like we need to add a lot of code to make it work for platforms that cannot run prover at all. This makes very little sense. 

I think we can try adding the stubs for non-x86_64 in _this_ module (which would probably more reasonable), but try to make them minimal.

ia-agentic: Changed it in that direction. I removed the interpreter fallback and kept the existing x86_64 JIT flow intact. On non-x86_64 targets, `gpu_prover/src/execution/simulation_runner.rs` now defines only minimal local `Context`/`JittedCode` stubs so `gpu_prover` still compiles with `ZKSYNC_USE_CUDA_STUBS=1`, but execution will panic if it actually reaches the simulator. Verified with `ZKSYNC_USE_CUDA_STUBS=1 cargo check -p gpu_prover` and `ZKSYNC_USE_CUDA_STUBS=1 cargo check -p gpu_prover --target aarch64-unknown-linux-gnu`.

gpu_prover/src/execution/simulation_runner.rs
```diff
@@ -24,6 +27,42 @@ use std::sync::{Arc, Mutex};
...
+
+    fn run_over_prepared_memory(
+        &self,
+        _context: &mut Context<I>,
+        _memory: &mut MemoryHolder,
+        _initial_trace_chunk: NonNull<TraceChunk>,
+    ) {
+        panic!("gpu_prover simulation requires the x86_64 JIT backend");
+    }
+}
```
popzxc: Can we move that to an inline module, e.g. `mod compat`, and then have two lines nearby:

```rust
#[cfg(target_arch = "x86_64")]
use riscv_transpiler::jit::{Context, JittedCode};
#[cfg(not(target_arch = "x86_64"))]
use self::compat::{Context, JittedCode};
```

This way I think it will be the most maintainable and readable.

gpu_prover/src/execution/simulation_runner.rs
```diff
@@ -227,7 +266,7 @@ impl<ND: NonDeterminismCSRSource + Send + 'static, T: TracingType + 'static>
         let snapshot_index = self.snapshot_index;
         self.snapshot_index += 1;
         let mut machine_state = *machine_state;
-        let timestamp = machine_state.timestamp.next_multiple_of(TIMESTAMP_STEP); // align timestamp, needs to be fixed in the VM
+        let timestamp = machine_state.timestamp.next_multiple_of(TIMESTAMP_STEP);
```
popzxc: Please do not change unrelated code, keep it as is.

gpu_prover/src/execution/simulation_runner.rs
ia-agentic: Moved the non-`x86_64` stubs into a local `compat` module and switched the imports to the cfg-gated `use` pattern you suggested in 10e938d7.

gpu_prover/src/execution/simulation_runner.rs
ia-agentic: Restored the unrelated timestamp comment in 10e938d7 so this follow-up only contains the requested `compat`-module cleanup.

gpu_prover/src/execution/simulation_runner.rs
```diff
@@ -10,9 +12,10 @@ use era_cudart_sys::{cudaHostRegister, cudaHostUnregister};
 use itertools::Itertools;
 use log::{debug, trace};
 use riscv_transpiler::common_constants::ROM_WORD_SIZE;
+#[cfg(target_arch = "x86_64")]
+use riscv_transpiler::jit::{Context, JittedCode};
```
popzxc: Please put these blocks together, with empty lines before and after, and right above `compat` module, e.g.

```
// ...imports

#[cfg(not(target_arch = "x86_64"))]
use self::compat::{Context, JittedCode};
#[cfg(target_arch = "x86_64")]
use riscv_transpiler::jit::{Context, JittedCode};

#[cfg(not(target_arch = "x86_64"))]
mod compat {
// ...
```

This way rustfmt shouldn't bring it together with other imports, and this chunk will clearly be seen as alternate import paths. 

Do not forget to run cargo fmt after that.

gpu_prover/src/execution/simulation_runner.rs
```diff
@@ -24,6 +27,42 @@ use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 use type_map::concurrent::TypeMap;
 
+#[cfg(not(target_arch = "x86_64"))]
+mod compat {
```
popzxc, line 36: Add a comment to this module stating the following:

```rust
// We're depending on JIT unconditionally, so it can fail compilation on some platforms,
// since some of the exported paths are platform-dependent. So, we provide panicking
// replacements to allow compilation on the platforms, e.g. with
// `ZKSYNC_USE_CUDA_STUBS`.
```

gpu_prover/src/execution/simulation_runner.rs
ia-agentic: Moved the cfg-gated `use` block so it now sits directly above `compat` with blank lines around it, and ran `cargo fmt` in 119f8275.

gpu_prover/src/execution/simulation_runner.rs
ia-agentic: Added that module comment verbatim in 119f8275. Rechecked with `ZKSYNC_USE_CUDA_STUBS=1 cargo check -p gpu_prover` and `ZKSYNC_USE_CUDA_STUBS=1 cargo check -p gpu_prover --target aarch64-unknown-linux-gnu`.

popzxc (approved):

robik75 (approved):

PR #251, opened 2026-03-30
feat(gkr): add word-only and subword-only memory circuits

mabbamOG:
## Summary
- add `mem_word_only` and `mem_subword_only` GKR circuits
- add and wire the RAM/ROM tables needed for subword memory ops
- regenerate the corresponding compiled circuit layouts and witness-eval artifacts

## Notes
- both memory GKR circuits are now working end-to-end


prover/src/gkr/sumcheck/evaluation_kernels/kernel_impls/max_quadratic_rel.rs
```diff
@@ -165,8 +165,11 @@ impl<F: PrimeField, E: FieldExtension<F> + Field>
                 #[cfg(feature = "gkr_self_checks")]
                 assert!(*a < input.len());
 
-                let mut contribution = *input.get_unchecked(*a);
                 for (b, coeff) in set.iter() {
+                    // Each (b, coeff) term is an independent contribution that starts from
+                    // input[a]. Reusing the previous contribution would incorrectly compound
+                    // quadratic terms.
+                    let mut contribution = *input.get_unchecked(*a);
```
shamatar: It's memory read, so we should move it out of cycle, but restructure the computation. Will fix myself

prover/src/gkr/prover/setup.rs
```diff
@@ -97,6 +97,9 @@ impl<F: PrimeField + TwoAdicField> GKRSetup<F> {
                 for column in width..padding_len {
                     result[column][row_idx + offset] = F::ZERO;
                 }
+                for column in width..padding_len {
```
shamatar: Merge error

mabbamOG:
## Latest changes (3ca4aa7d)

**Optimizations to both mem circuits:**
- Replace `RomAddressSpaceSeparator` lookup table with arithmetic isrom detection (range check on residue), saving 1 lookup + 1 variable per circuit
- Subword: total_tables 14,352,385 → 14,286,849 (-65,536), generic_lookups 4 → 3
- Word-only: total_tables 2,162,689 → 2,097,153 (-65,536), generic_lookups 3 → 2

**Word-only circuit additionally:**
- Replace explicit addr/romread variables with implicit address computation from memory accesses + single merged final lookup (matching subword pattern)

**Compiler fix:**
- Fix `layout_lookup_expressions` loop exiting early when generic lookups exist only at higher GKR layers (no lookups at intermediate layers to keep the loop alive)

**New API:**
- `Circuit::require_invariant_from_lookup_input` for range-checking `LookupInput` expressions (not just `Variable`s)

All cs/witness_eval_generator/prover tests pass (prover tested with `--release`).

shamatar (approved):

PR #258, opened 2026-04-10
fix: missing checks in verifiers after refactoring

shamatar:
## What ❔

Previously continuity of challenges was checked by verifier function pointers, but now it's not, even though comments told otherwise.

## Is this a breaking change?
- [ ] Yes
- [x] No


full_statement_verifier/src/unrolled_proof_statement.rs
```diff
@@ -314,18 +321,22 @@ pub unsafe fn verify_full_statement_for_unrolled_circuits<
...
 
-                    // now we should check all invariants about continuity
-
                     delegation_used |= true;
 
+                    // now we should check all invariants about continuity, and it's a circuit type that
+                    // uses delegation argument and outputs it. So we compare all challenges across circuit
+                    // instances with delegations (of only single type!), and after the cycle compare between
+                    // circuits with different ProofOutput types
                     if circuit_sequence > 0 {
```
olesHolem, line 336: Should we also switch this to `if total_cycles > 0` for challanges check, or it there a guarantee that UnrolledWithDelegation comes first?

olesHolem: It seems like `SHIFT_BINARY_CSR_VERIFIER_PTR` is the only `UnrolledWithDelegation` and it comes after `ADD_SUB_LUI_AUIPC_MOP_VERIFIER_PTR`

full_statement_verifier/src/unrolled_proof_statement.rs
```diff
@@ -340,15 +351,44 @@ pub unsafe fn verify_full_statement_for_unrolled_circuits<
...
     }
 
+    // Check that we actually run something meaningful
+    assert!(total_cycles > 0);
+
+    // Check that we have values initialized for challenge comparisons used below
+    assert!(non_delegation_used);
+
+    // Compare across delegation and non-delegation circuit types that challenges are the same
+    if delegation_used {
```
olesHolem: Oh, I see, here's this check

olesHolem (dismissed):

olesHolem (approved):

PR #267, opened 2026-04-22
feat: Unified verifiers security via const generics

popzxc:
Not ready for review yet, highly experimental


shamatar:
Fine by me in general, and I hope we will do better in the next version without GCE all together

Fitznik:
Can you clarify the purpose of this PR? Is the goal to preserve both options for cost/performance reasons?

Why can’t we completely remove the 80-bit verifier? I understand that keeping it may be cheaper, but then we are missing the point of targeting 100-bit security. I’d also prefer not to overcomplicate the system by maintaining two versions of the circuits, two versions of the verifier, and two versions of the transpilers.

popzxc:
@Fitznik there might be a transitional period where some of the networks still use 80bit security, some use 100bit security, and if we drop 80bit security support, it means that networks that still use 80bit cannot benefit from patches & updates. Imagine that we sill use 80bit somewhere, and there is a security patch. We now will have to introduce a branch from the last commit with 80bit support and make a release from there, and probably maintain it in case we'll need it again.

It's better to drop support for 80bit when we know for sure that there are no clients that use 80bit security anywhere.

Otherwise I agree that it's not a long-term solution.

Fitznik (approved):
Looks good to me. I ran the related tests and the CPU verifier paths seem to be working. You probably already know this, but just in case: we should also sync zkos-wrapper with these Airbender changes.

PR #270, opened 2026-04-23
fix: remove generic const expression in verifier

shamatar:
## What ❔

Can now build the verifier on stable

## Is this a breaking change?
- [ ] Yes
- [x] No


verifier/src/lib.rs
```diff
@@ -1,9 +1,6 @@
 #![cfg_attr(not(any(test, feature = "replace_csr")), no_std)]
-#![allow(incomplete_features)]
-#![feature(generic_const_exprs)]
-
-#[cfg(any(test, feature = "proof_utils"))]
-extern crate alloc;
+#![cfg_attr(not(any(test, feature = "proof_utils")), allow(incomplete_features))]
+#![cfg_attr(not(any(test, feature = "proof_utils")), feature(generic_const_exprs))]
```
yoaveshel: Why are these opposite from `verifier_common`? (`not(any(test, feature = "proof_utils"))` vs `any(test, feature = "proof_utils")`)

transcript/src/lib.rs
```diff
@@ -249,6 +249,27 @@ impl Blake2sTranscript {
...
+        need_hashing_round: bool,
+    ) {
+        unsafe {
+            if need_hashing_round {
+                Self::draw_randomness_inner(hasher, seed);
+            }
+
+            let mut dst_ptr: *mut u32 = dst.as_mut_ptr().cast::<u32>();
+            // first we can just take values from the seed
+            crate::spec_memcopy_u32_nonoverlapping(
```
yoaveshel: shouldn't we advance the seed here?

verifier/src/lib.rs
shamatar: True

transcript/src/lib.rs
shamatar: When we have a seed, we first “use” the seed, and when it’s exhausted - we hash the seed and get a new one, and then use new one, and so on

shamatar: Oh, I forgot to delete this function all together after changing an iterator. Ignore it. Looks like I missed dead code lint

PR #277, opened 2026-04-27
gkr verifier stats

mabbamOG:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


field/src/baby_bear/ext4.rs
```diff
@@ -155,6 +159,10 @@ impl Field for BabyBearExt4 {
 
     #[cfg_attr(not(feature = "no_inline"), inline(always))]
     fn sub_assign(&'_ mut self, other: &Self) -> &'_ mut Self {
+        #[cfg(feature = "verifier_stats")]
+        {
+            crate::stats::FIELD_STATS.borrow_mut().fext_adds += 1;
```
shamatar: Do you have a mechanism somewhere to correct number of base field ops due to extension field ops? Here you will get 4 base field additions added to the counter recursively

field/src/lib.rs
```diff
@@ -2,6 +2,10 @@
 #![cfg_attr(target_arch = "riscv32", allow(internal_features))]
 #![cfg_attr(target_arch = "riscv32", feature(core_intrinsics))]
 #![cfg_attr(target_arch = "riscv32", feature(const_eval_select))]
+#![cfg_attr(
```
shamatar: You can just do
```
#![cfg_attr(not(any(test, feature = "verifier_stats")), no_std)]
```

and avoid using unstable feature attribute. Same for other crates. So far our verifier can be built "almost on stable" with only `const_eval_select` being an outlier, and we want to keep it this way

field/src/baby_bear/ext4.rs
mabbamOG: ext2 ops feed directly into _impl chain which has no counters.
the wrapping at high level from op -> op_impl is great for separation of counters.

mabbamOG:
fixed issues @shamatar

common_constants/src/stats.rs
```diff
@@ -0,0 +1,14 @@
+#[cfg(feature = "verifier_stats")]
+extern crate std;
```
shamatar, line 2: I didn't know that you can do such trick not at the crate top, but at the module level. Nice one

shamatar (approved):

common_constants/src/stats.rs
mabbamOG: i did not know either, the machine came up with it. barely any point coming up with these things by hand anymore imo

PR #280, opened 2026-04-28
feat: Support benchmarking CPU parts of GPU prover

popzxc:
CPU often becomes a bottleneck when running GPU prover, so this PR makes it possible to independently benchmark CPU parts only.

It is possible to make diff smaller by using features instead of generics, but we have so many features already that I'm hesitant to add more.

The added bench can be ran as `cargo bench -p gpu_prover --bench cpu_pipeline`.

### DIFF
--- a/gpu_prover/src/execution/cpu_pipeline_model.rs
+++ b/gpu_prover/src/execution/cpu_pipeline_model.rs
@@ -1,0 +1,727 @@
+use super::cpu_worker::{collect_inits_and_teardowns, get_inits_and_teardowns_chunks};
+use super::messages::{InitsAndTeardownsData, SimulationResult, TracingData, WorkerResult};
+use super::simulation_runner::{
+    LockedBoxedMemoryHolder, LockedBoxedTraceChunk, SimulationRunner, Snapshot,
+};
+use super::tracing::{SplitTracingType, Tracer, TracingType, UnifiedTracingType};
+use super::A;
+use crate::circuit_type::{CircuitType, UnrolledCircuitType};
+use crate::execution::prover::ExecutionKind;
+use crate::machine_type::MachineType;
+use crate::prover::tracing_data::{
+    DelegationTracingDataHost, TracingDataHost, UnrolledTracingDataHost,
+};
+use crate::witness::trace_unrolled::ShuffleRamInitsAndTeardownsHost;
+use crossbeam_channel::{unbounded, Receiver, Sender};
+use cs::definitions::TimestampScalar;
+use era_cudart::memory::{CudaHostAllocFlags, HostAllocation};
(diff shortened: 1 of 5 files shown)


gpu_prover/src/execution/cpu_pipeline_model.rs
```diff
@@ -0,0 +1,770 @@
...
+    timings.replayed_cycles += total_cycles as u64;
+}
+
+fn recycle_snapshots<T: TracingType>(
+    snapshots: Receiver<Snapshot<T::Ranges, T::Allocator>>,
+    free_trace_chunks: Sender<LockedBoxedTraceChunk<T::Allocator>>,
+    results: Sender<WorkerResult<T::Allocator>>,
+) {
+    for snapshot in snapshots {
+        let Snapshot { index, trace, .. } = snapshot;
```
robik75: I found a SimulationOnly ownership race that can panic the new bench.

recycle_snapshots destructures snapshots as:

```
let Snapshot { index, trace, .. } = snapshot;
free_trace_chunks.send(trace).unwrap();
results.send(WorkerResult::SnapshotReplayed(index)).unwrap();
```

The omitted trace_ranges field is not dropped immediately; it stays alive until the end of the loop body. Those ranges hold `PtrRange::_chunk Arc<Vec<_>>` clones, while the produced `TracingData` owns the same chunks. Once `SnapshotReplayed` is sent, the main thread may free the matching tracing data and call `Arc::into_inner(...).unwrap()`, but the recycler thread can still be holding the other Arc, causing:

thread 'main' panicked at gpu_prover/src/witness/trace.rs:32:41:
called `Option::unwrap()` on a `None` value
Repro:

`RUST_BACKTRACE=1 cargo bench -p gpu_prover --bench cpu_pipeline -- SimulationOnly/replay_threads=1`

It may take more than one run because the panic depends on scheduling.

A minimal fix is to explicitly drop trace_ranges before reporting the snapshot as replayed/recycled:

```
let Snapshot {
    index,
    trace,
    trace_ranges,
    ..
} = snapshot;
free_trace_chunks.send(trace).unwrap();
drop(trace_ranges);
results.send(WorkerResult::SnapshotReplayed(index)).unwrap();
```

I tested this locally and the same filtered benchmark completed successfully.

robik75: Actually cpu_worker.rs (line 209) seems to have the same semantic ordering problem in production replay: tracer owns trace_ranges, but SnapshotReplayed is sent at line 230 before tracer is dropped at end of scope. This is pre-existing on dev branch.
I prepared a fix PR #283

robik75 (changes requested):

robik75 (approved):

PR #281, opened 2026-04-28
feat: unreduced accumulation

yoaveshel:
## What ❔
Proof of concept for unreduced accumulation in prover. Applied to step 0 of batched sumcheck and resulted in a ~2.1s reduction in prover time for add sub (without self-checks). Benches show 3x improvement compared to Montgomery reduction at every op. 

To integrate this to later steps we need to implement the split eq and small value optimisations from https://eprint.iacr.org/2026/587.pdf

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/field/src/baby_bear/unreduced.rs
+++ b/field/src/baby_bear/unreduced.rs
@@ -1,6 +1,7 @@
 use crate::baby_bear::base::BabyBearField;
+use crate::baby_bear::ext2::BabyBearExt2;
 use crate::baby_bear::ext4::BabyBearExt4;
-use crate::field::{FieldExtension, UnreducedAccumulator};
+use crate::field::UnreducedAccumulator;
 
 #[derive(Clone, Copy, Debug)]
 #[repr(transparent)]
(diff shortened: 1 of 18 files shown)


field/src/baby_bear/unreduced.rs
```diff
@@ -0,0 +1,248 @@
...
+    #[cfg_attr(not(feature = "no_inline"), inline(always))]
+    pub fn add_assign_ext(&mut self, e: BabyBearExt4) {
+        let coeffs = <BabyBearExt4 as FieldExtension<BabyBearField>>::into_coeffs(e);
+        for k in 0..4 {
+            let delta = (coeffs[k].0 as u128) << 32;
+            self.lanes[k] = self.lanes[k].wrapping_add(delta);
+        }
+    }
+
+    pub fn finalize(self) -> BabyBearExt4 {
```
shamatar, line 84: You are working on the concrete type with internal fields being available, so it's just easier to use an explicit constructor. Also do not forget inline

field/src/baby_bear/unreduced.rs
```diff
@@ -0,0 +1,248 @@
...
+            BabyBearRawProductSum(l1).finalize(),
+            BabyBearRawProductSum(l2).finalize(),
+            BabyBearRawProductSum(l3).finalize(),
+        ];
+        <BabyBearExt4 as FieldExtension<BabyBearField>>::from_coeffs(coeffs)
+    }
+}
+
+impl UnreducedAccumulator<BabyBearField, BabyBearExt4> for BabyBearExt4RawProductSum {
+    const ZERO: Self = Self::ZERO;
```
shamatar, line 100: If you want to micro-optimize, then you should check the assembly to see it a sequence of making 0 accumulator and then adding first pair is optimized out by llvm. If not - most likely you want to add a constructor from pair itself, and a function for fixed-sized loops (where loop/input size is const generic) where first iteration is unrolled as constructor and loop size is asserted >0

PR #282, opened 2026-04-29
fix: add virtual poly check in verifier

yoaveshel:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


verifier/src/generated/jump_branch_slt/gkr.rs
```diff
@@ -3812,6 +3812,52 @@ pub(crate) fn verify_gkr<I: NonDeterminismSource, E: ErrorCreator>(
                 return Err(E::gkr_lookup_identity_failed(2usize));
             }
         }
+        unsafe {
```
shamatar: Is there any way to add comments in the generated token stream? I know what to expect here, but for unprepared reader it's better to refer to https://github.com/matter-labs/zksync-airbender/blob/av_gkr_compiler/prover/src/gkr/virtual_polys/range_check.rs here

verifier/src/generated/add_sub_lui_auipc_mop/gkr.rs
```diff
@@ -3854,6 +3854,52 @@ pub(crate) fn verify_gkr<I: NonDeterminismSource, E: ErrorCreator>(
...
+                k += 1;
+            }
+            while k < 24usize {
+                let mut t: BabyBearExt4 = BabyBearExt4::ONE;
+                let p = pt.get_unchecked(24usize - 1 - k);
+                field_ops::sub_assign(&mut t, &*p);
+                field_ops::mul_assign(&mut result, &t);
+                k += 1;
+            }
+            if result != *state.prev_claims.get_unchecked(55usize) {
```
shamatar: I do not understand where this index comes from. Here https://github.com/matter-labs/zksync-airbender/blob/av_gkr_compiler/verifier/src/generated/add_sub_lui_auipc_mop/constants.rs `LAYER_0_SORTED_ADDRS` has virtual setups at index 39 and 40

verifier/src/generated/jump_branch_slt/gkr.rs
yoaveshel: doc comments survive the lexer as `#[doc = ".."]`

verifier/src/generated/add_sub_lui_auipc_mop/gkr.rs
yoaveshel: it's into `state.prev_claim` which is merged and in a different order. `LAYER_0_SORTED_ADDRS` is dead code, I'll remove it

verifier/src/generated/add_sub_lui_auipc_mop/constants.rs
shamatar: I think we need to generate another "dead" constant that is a mapping of index -> GKRAddress. And it's still not clear how openings are done - below there are 55 claims. `ORACLE_NUM_COLS` has 55 columns in total. And in the comment below for `check_virtual_setup_range_check_16bits` you refer to `INITIAL_WHIR_CLAIM_INDICES`. But claims for virtual polys must not participate in WHIR (they are not in the setups)

shamatar (changes requested):

verifier/src/generated/add_sub_lui_auipc_mop/constants.rs
yoaveshel: the "dead" constant is purely for documentation?
And the comment is misleading, I'll fix it. Whir only reads 0..55, virtual polys live in 55, 56 which whir doesn't touch

shamatar: You have a readme below
```
#[doc = " The `prev_claims` index is the position in this layer's merged target_addrs (regular + cache-input extras, BTreeSet-sorted),"]
#[doc = " which is also the layout used by `INITIAL_WHIR_CLAIM_INDICES`."]
```
and use 55 and 56 indexes for virtual poly evals, but they are not present in this array. So I'm asking if we should make an array (for documentation purposes only) that is a effectively a map of index -> poly for `*state.prev_claims.get_unchecked(56usize)` like accesses at base layer

yoaveshel: ok I cleaned up the indexing and removed the indirection - so hopefully everything is clearer now

PR #298, opened 2026-05-15
[WIP] feat: Expr IR

popzxc:
Prototype implementation for expressions IR.


mabbamOG:
what is the need for Scale when Product is already there? just to save some ast traversal?

popzxc:
> what is the need for Scale when Product is already there? just to save some ast traversal?

Mostly yes. It's not needed technically, but I thought that it's more intuitive and not really harmful. Is there a reason to keep the AST minimal?

It also potentially reduces the number of allocations as well (since we won't need to use `Vec` each time), which is a nice property IMHO.

PR #299, opened 2026-05-15
perf(field): shorten BabyBear inverse chain from 40 to 37 ops

robik75:
## What ❔

Shortens the BabyBear Fermat-inverse chain in `field/src/baby_bear/base.rs` from **30 sqr + 10 mul (40 ops)** to **29 sqr + 8 mul (37 ops)** via a 6-bit-window addition chain: build `x^119` (top 7 bits of `p-2`, which is `0b1110` followed by 27 ones), then four rounds of `(sqr^6, mul x^63)` fill the remaining 24 bits. Also updates the `verifier_stats` counter from `40` to `37`.

## Why ❔

`Field::inverse` is called from the prover, `cs`, and host-side tooling. The riscv32 verifier does not invert at runtime — its inverse powers are precomputed and baked into the proof as constants — so this is a no-op there. Algorithm-only change; bit-for-bit equivalent to `pow(p-2)`.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

- [x] PR title corresponds to the body of PR.
- [x] Tests for the changes have been added / updated. (`test_inversion_chain` and the `mul_inverse` proptest already cover this.)

### DIFF
--- a/field/src/baby_bear/base.rs
+++ b/field/src/baby_bear/base.rs
@@ -140,8 +140,8 @@ impl BabyBearField {
     }
 
     pub(crate) const fn inverse_impl(&self) -> Option<Self> {
-        // a^(p-2) - it's anyway expensive operation, but on platform with
-        // field ops it is still faster than binary GCD
+        // a^(p-2) — Fermat's little theorem. Faster than binary GCD on platforms
+        // with native modular multiplication.
 
         if self.is_zero_impl() {
             return None;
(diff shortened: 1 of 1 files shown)


yoaveshel (approved):
We could add a proptest to check `x.inverse() == x.pow(CHAR - 2)` but other than that lgtm

robik75:
> We could add a proptest to check `x.inverse() == x.pow(CHAR - 2)` but other than that lgtm

done

PR #309, opened 2026-06-09
fix: bugs in add-sub circuit

yoaveshel:
## What ❔
Fix 3 bugs in add sub circuit:
- mulmod / fmamod missing R^-1 factor
- fmamod not included in canonicity check
- wrong sign in modular high limb reduction

Previously our test binaries didn't include mop instructions so these bugs were missed. This PR adds a binary exercising all mop instruction and prover/verifier tests that run it.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


cs/src/gkr_circuits/utils.rs
```diff
@@ -56,6 +56,20 @@ pub fn calculate_pc_next_no_overflows_with_range_checks<F: PrimeField, CS: Circu
     circuit.add_constraint_allow_explicit_linear_prevent_optimizations_expr(pc_high);
 }
 
+pub(crate) fn montgomery_product<F: PrimeField, W: WitnessTypeSet<F>>(
+    a: &W::Field,
+    b: &W::Field,
+) -> W::Field {
+    let mut product = a.clone();
```
shamatar: I do not think it's a valid fix. We want to have roundtrip u32 -> field interpretation of this value (so "from_raw_repr" function), then just normal multiplication (that for Montgomery form would be different than e.g. for M31 field, but still self-consistent), and then back to u32 raw representation. So think you just need to fix WitnessComputationalField only, and not this part

cs/src/gkr_circuits/utils.rs
```diff
@@ -56,6 +56,20 @@ pub fn calculate_pc_next_no_overflows_with_range_checks<F: PrimeField, CS: Circu
...
+    product.mul_assign(b);
+    product.mul_assign(&W::Field::constant(F::from_reduced_raw_repr(1)));
+    product
+}
+
+pub(crate) fn montgomery_product_expr<F: PrimeField>(a: Expr<F>, b: Expr<F>) -> Expr<F> {
+    a * b * Expr::<F>::constant(F::from_reduced_raw_repr(1))
```
shamatar: And here you should branch of F::MONT_REPR boolean and either use or not extra prefactor

shamatar (changes requested):

cs/src/gkr_circuits/utils.rs
yoaveshel: We still need the R^-1 factor in the constraint, no?

yoaveshel: for the witness I did as you asked

shamatar:
I still do not get why you need new functions to cast from and into raw repr. Just old functions from and into integer should just use from/as raw repr instead. We always want to work with the representation that is native to the field in risc-v, same as for native prover. So if integer representation of field element is Xu32 in register, then for witness evaluation we just want to cast it into F(Xu32), and if eventually we will get a Mont form product as F(Yu32), then we would want to put into register just Yu32. So the product constraint that depends on whether field is Montgomery form or not is correct, but not the new cast function

cs/src/witness_placer/graph_description/field.rs
```diff
@@ -239,4 +239,15 @@ impl<F: PrimeField> WitnessComputationalField<F> for FieldNodeExpression<F> {
     fn from_integer(value: Self::IntegerRepresentation) -> Self {
         Self::FromInteger(Box::new(value))
     }
+    fn from_raw_repr_with_reduction(value: Self::IntegerRepresentation) -> Self {
+        let mut x = Self::from_integer(value);
```
shamatar: Here it should be just F::from_raw_repr

cs/src/witness_placer/graph_description/field.rs
```diff
@@ -239,4 +239,15 @@ impl<F: PrimeField> WitnessComputationalField<F> for FieldNodeExpression<F> {
...
+        x.mul_assign(&Self::constant(F::from_reduced_raw_repr(1)));
+        x
+    }
+    fn into_raw_repr_reduced(self) -> Self::IntegerRepresentation {
+        let r = F::from_reduced_raw_repr(1).inverse().unwrap();
```
shamatar: And here too

cs/src/witness_placer/graph_description/field.rs
shamatar: Oh, for the expression - it should be just new expression type and witness evaluation generator should handle it

shamatar (approved):

PR #310, opened 2026-06-09
fix!: LHU memory lookup  

Fitznik:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/cs/src/tables/memory_opcode_related.rs
+++ b/cs/src/tables/memory_opcode_related.rs
@@ -215,7 +215,7 @@ pub fn create_store_byte_existing_contribution_table<F: PrimeField>(id: u32) -> 
 }
 
 pub fn create_memory_offset_mask_with_trap_table<F: PrimeField>(id: u32) -> LookupTable<F, 3> {
-    const TABLE_MAX_WIDTH: usize = 16 + 3 + 1 + 1; // offset low || funct3 || is_store || rd_is_zero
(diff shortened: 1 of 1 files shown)


cs/src/tables/memory_opcode_related.rs
```diff
@@ -229,8 +229,8 @@ pub fn create_memory_offset_mask_with_trap_table<F: PrimeField>(id: u32) -> Look
 
             let mem_address_low = input & 0xffff;
             let funct3 = (input >> 16) & 0b111;
-            let is_store = (input >> 17) & 0b1 != 0;
-            let rd_is_x0 = (input >> 18) & 0b1 != 0;
+            let is_load = (input >> 19) & 0b1 != 0;
```
shamatar, line 232: I do not think you can change the meaning of this bit without changing a circuit. Or at least confirming that circuit actually puts "load" Boolean there

Fitznik: I made this fix to mirror the circuit behavior. https://github.com/matter-labs/zksync-airbender/blob/c5dc66c57266e5d2b0f5b8a9d5035f107e92077e/cs/src/machine/ops/unrolled/load_store_subword_only.rs#L74

shamatar: Indeed, I thought that table at least interprets the bit with a proper semantics, but no

shamatar (approved):

PR #305, opened 2026-05-21
feat: add gkr unified circuit

yoaveshel:
## What ❔

- New unified circuit (113 columns = 42 memory + 61 witness + 10 setup) + decoder trait infrastructure
- Cross-family resource sharing: a shared base-layer scratch pool (bools + vars + RC-16 register, sized max across families, aliased per-row since dispatch is one-hot) + pooled lookups, gated by family exec masks. Enabled the SINGLE_COLUMN range-check merge node. Cuts witness columns 73→61.
- 21 constraint-level negative tests + 4 proof-level (real-verifier) rejection tests + an inits/teardowns eval-corruption test; SLT/BLT/BGE table regression tests

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] Code has been formatted.

PR #318, opened 2026-06-23
feat: add unified full statement verifier

yoaveshel:
## What ❔
Gated on https://github.com/matter-labs/zksync-airbender/pull/305

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


full_statement_verifier/src/unified_circuit_statement.rs
```diff
@@ -61,242 +64,159 @@ pub unsafe fn verify_unified_circuit_statement<const BASE_LAYER: bool>(
...
 
-    // loop over main circuit type
-    let mut proof_output_0: ProofOutput<CAP_SIZE, NUM_COSETS, NUM_DELEGATION_CHALLENGES, 1, 1> =
-        MaybeUninit::uninit().assume_init();
-    let mut proof_output_1: ProofOutput<CAP_SIZE, NUM_COSETS, NUM_DELEGATION_CHALLENGES, 1, 1> =
-        MaybeUninit::uninit().assume_init();
-    let mut state_variables = ProofPublicInputs::uninit();
+    // single reduced-machine family with folded inits/teardowns
+    let num_unified_circuits = nd_source.read_word();
+    assert!(num_unified_circuits == 1);
```
shamatar: We can not assume it in general - it may be better to switch to unified circuits even when we get >1 of them just because proving unrolled ones doesn't fill them well enough for all types

full_statement_verifier/src/unified_circuit_statement.rs
```diff
@@ -61,242 +64,159 @@ pub unsafe fn verify_unified_circuit_statement<const BASE_LAYER: bool>(
...
+        buffer[0] = common_constants::REDUCED_MACHINE_CIRCUIT_FAMILY_IDX as u32;
+        transcript.absorb(&buffer);
+    }
 
     let mut total_cycles = 0u64;
-    for circuit_sequence in 0..num_circuits {
-        total_cycles += unified_circuit_capacity as u64;
+    // strictly-increasing check over the concatenated `top_bits` of all i/t-carrying
+    // instances: every covered prefix is unique, so no two instances init the same range.
+    let mut prev_top_bit: i64 = -1;
```
shamatar: You can do i32 here I think

full_statement_verifier/src/unified_circuit_statement.rs
```diff
@@ -61,242 +64,159 @@ pub unsafe fn verify_unified_circuit_statement<const BASE_LAYER: bool>(
...
-    // since we have > 0 main circuits, then we can always use `proof_output_0` below
-
-    // ok, now we forget about main circuit and potentially parse delegations
-    if NUM_DELEGATION_CHALLENGES > 0 {
-        let mut previous_delegation_type = 0u32;
-        let mut state_variables = ProofPublicInputs::uninit();
-        let mut delegation_proof_output = MaybeUninit::uninit().assume_init();
-
-        let mut total_delegation_requests = 0u64;
+    let mut total_permutation_elements = total_cycles << 2; // 4 permutation elements per cycle
```
shamatar, line 176: Can you check that we didn't miss similar shift in unrolled verifier, and then
```
total_permutation_elements +=
                    delegation_circuit_params.num_permutation_terms_per_circuit as u64;
```
line?

full_statement_verifier/src/unified_circuit_statement.rs
yoaveshel: Do we have a way now to get a multi-instance proof? If I'm enabling support for it (which requires some generator changes) I'd like to add a test for it as well

shamatar (changes requested):
only >1 assumption is essential and important change, rest are smaller details

full_statement_verifier/src/unified_circuit_statement.rs
```diff
@@ -61,242 +64,159 @@ pub unsafe fn verify_unified_circuit_statement<const BASE_LAYER: bool>(
...
+            let it = proof_output
+                .inits_and_teardowns
+                .expect("trailing unified instance must carry inits/teardowns");
+            read_set_product_accumulator.mul_assign(&it.read_product);
+            write_set_product_accumulator.mul_assign(&it.write_product);
+
+            for top_bit in proof_output.inits_and_teardowns_top_bits.iter() {
+                assert!((*top_bit as i64) > prev_top_bit);
+                prev_top_bit = *top_bit as i64;
             }
```
shamatar, line 167: I also think we need "else" branch here, that will assert that all init/teardown top bits used by the prover are 0. It'll remove any "free will" for circuits where we ignore inits/teardowns on their contribution to FS transformation eventually

PR #330, opened 2026-06-29
feat: derive memory pow bits

yoaveshel:
## What ❔
Replace the inert `MEMORY_DELEGATION_POW_BITS = 0; // TODO` with a derived, single-sourced PoW-bit count gating the memory/delegation permutation-argument challenges.

### Changes

- `verifier_common`: derive the `count — base = 123 − 40 − 2 = 81`, `bits = max(0, level−base)` → 80→0, 100→19; `MAX_PERMUTATION_ELEMENTS_LOG2` derived from the RAM timestamp constants; `security_80/100` made mutually exclusive (`compile_error!`).
- `full_statement_verifier`: re-export the constant; couple the total_permutation_elements assert to it; forward security features.
- `gkr_test.sh`: dropped `--security-level both`; native step runs the self-check at both levels.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


verifier_common/src/lib.rs
```diff
@@ -39,6 +39,133 @@ pub const fn transcript_challenge_array_size(num_elements: usize, pow_bits: usiz
...
+/// `full_statement_verifier::MAX_CYCLES` computes it; the delegation headroom adds the policy
+/// margin. Currently `36 + 2 + 2 = 40`.
+///
+/// The coupling is structural, not by convention: BOTH the runtime assert in
+/// `full_statement_verifier::unrolled_proof_statement` AND the PoW derivation
+/// ([`memory_delegation_pow_bits`]) read this one constant, so they cannot drift — changing the
+/// cycle layout re-derives the PoW automatically. The `derivation_matches_expected_values`
+/// self-check pins the resulting value/base soundness as a tripwire: a layout change that moves
+/// it fails the test and forces a deliberate PoW-soundness re-review.
+pub const MAX_PERMUTATION_ELEMENTS_LOG2: usize = {
```
shamatar: There is no logical link to be honest between max number of cycles that risc-v circuits can have (that one is (2^38 - initial) / 4) and how many elements we accept to see in total (as if we have a lot of delegations then those add quite a few elements to permutations, but do not drive timestamps). So here we can just hardcode 40 as some sane limit (and hardly reachable in practice) just for purposes of PoW derivation

shamatar: if e.g. for some security level our PoW would be too much, we could reduce this value instead. But so far we do not need it

shamatar (changes requested):

PR #332, opened 2026-07-02
feat: add tri add and xor rot opcode to unified circuit

yoaveshel:
## What ❔
This PR also brings in av_convergence_test

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


cs/src/gkr_circuits/unified_reduced_machine/binary_shifts.rs
```diff
@@ -55,6 +58,10 @@ pub fn apply_unified_binary_shifts_inner<F: PrimeField, CS: Circuit<F>>(
     let rs1_b2 = scratch_space[18];
     let rs2_b0 = scratch_space[19];
     let rs2_b2 = scratch_space[20];
+    // xor-rotate (unified-only) reads rd_old (rd_read_limbs) as the 2nd XOR operand; split its
```
shamatar: I see you decided to blindly use rd's value and decomposition. Why did you reject an option to remap rd -> rs2 (even though RV opcode format would formally has rs2 == 0 and imm != 0, in decoder you anyway set imm to 0, so you could just set rs2 to rd in riscv_transpiler crate and carelly propagate everything)

shamatar: Deeper change would be to just merge it with binary ops, saving also 1 more bit in decoder

shamatar (changes requested):

cs/src/gkr_circuits/unified_reduced_machine/add_sub_lui_auipc_mop.rs
```diff
@@ -441,6 +518,30 @@ pub fn apply_unified_add_sub_lui_auipc_mop_inner<F: PrimeField, CS: Circuit<F>>(
...
     }
 
+    // tri-add (unified-only): rd = rs1 + rs2 + rd_old (wrapping u32), per-16-bit-limb.
+    // Each limb carry ∈ {0,1,2} is encoded as a sum of two Booleans (chained-Boolean carries):
+    //   low  carry = intermediate_carry (of_slots[1]) + tri_clo_b (of_slots[2])
+    //   high carry = carry              (of_slots[0]) + tri_chi_b (of_slots[3])
+    // out_low/out_high are 16-bit range-checked (above) and rd_read limbs are 16-bit register
+    // reads, so each limb equation uniquely determines (out, carry); the discarded top carry
+    // (chi) ∈ {0,1,2} realises the mod-2^32 wrap. Both gated by is_tri_add (=0 ⇒ trivially sat).
+    {
```
shamatar, line 528: I'll accept a PR to unblock the work, but there will be few more efficiency comments. This constraint can be merged with the existing add/sub/lui/auipc I think

cs/src/gkr_circuits/unified_reduced_machine/binary_shifts.rs
```diff
@@ -90,9 +95,12 @@ pub fn apply_unified_binary_shifts_inner<F: PrimeField, CS: Circuit<F>>(
 
     let is_binary_op = decoder.perform_binary_op();
     let is_shift = decoder.perform_shift();
+    let is_xor_rot = xor_rot;
 
     {
-        let f3_sum: Constraint<F> = Term::from(is_binary_op) + Term::from(is_shift);
+        // At most one Family-3 sub-opcode fires per row (binary-op / shift / xor-rotate).
```
shamatar, line 101: This exclusivity is guaranteed by the decoder, so we do not need this constraint

shamatar (approved):

yoaveshel:
@shamatar ok just live the efficiency comments here so I can reference them and I'll address it in a follow-up

PR #340, opened 2026-07-11
feat: combine multiple recursion proofs into one in the unified layer

dimazhornyk:
## What ❔

Port of CombinedMultipleRecursionLayers from the old universal verifier (option 6, #141) to the unrolled/unified recursion design, enabling multi-batch proof aggregation before SNARKing.

Guest side: new OP_VERIFY_COMBINED_RECURSION_LAYERS_IN_UNIFIED_CIRCUIT = 3 handled by verify_combined_recursion_layers: verifies N >= 2 recursion layer proofs (each stream prefixed with its own unrolled/unified op; nested combining is rejected), asserts all proofs share the same recursion chain (output words 8..16), and returns keccak(out_1[0..8]>>32 || ... || out_n[0..8]>>32) in words 0..8 with the shared chain carried through - the exact hash format the old combined flow exposed to the SNARK wrapper.

Host side: CombinedRecursionInput, combined oracle flattening, expected output computation and native verification in execution_utils; new CLI `combine` command and ProofTarget::RecursionCombined with combine_artifacts, which verifies every input artifact, proves the combined statement with the unified-layer program, shrinks it to convergence via unified self-recursion, and self-checks the result against the expected rolling hash.

The unified-layer verifier binaries are rebuilt via the reproducible Docker flow (+~7.3KB); the unrolled-layer binaries are untouched.

Verified end to end on CPU (opcode_smoke, security 80): base -> unrolled ->
unified -> combine(proof, proof) -> verify; the combining run took 12.24M cycles for two proofs (fits the 2^24 smoke-test bound) and the final output matched an independently computed keccak rolling hash word for word. The cli_smoke_verify CI job now also exercises combine + verify.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


execution_utils/src/unrolled_gpu.rs
robik75, line 1: i don't think this code should live here, this is a specific variant of the gpu prover, i see it only used by the cli but i don't see it being used in the airbender-prover? not sure what the use is supposed to be outside of CLI, if it is only used by CLI then it should live there? CLI has a prover_utils module.

but then again maybe all these things do not belong to CLI at all if they are supposed to be reused, CLI is not supposed to be used as a library but i see it is used that way by airbender-prover while it is also using execution_utils. it has a separate `lib` in addition to `bin` which is weird on it's own.

this whole might need some rethinking/refactoring/change of architecture.

gpu_prover/src/prover/context.rs
```diff
@@ -48,15 +48,37 @@ impl Default for ProverContextConfig {
     fn default() -> Self {
         Self {
             powers_of_w_coarse_log_count: 12,
-            allocator_block_log_size: 20,             // 1 MB blocks
-            device_slack_static_bytes: 1 << 27,       // 128 MB static slack
-            device_slack_per_thread_bytes: 1 << 11,   // 2 KB per thread slack
+            allocator_block_log_size: 20, // 1 MB blocks
+            // The slack is VRAM left outside the pool for driver-side runtime needs
```
robik75: i smell Claude making made up claims, an L4 does not have  24GB of usable memory because of ECC, "prover needs ~23.2 GB" is nonsense
the issue with slack is probably true, just the numbers are not

anyway, these changes should not be here at all
do not add an environment variable to override default configuration, if you do not want default configuration, do not instantiate the prover with default configuration, the prover takes configuration as an argument for a reason

gpu_prover/src/prover/context.rs
```diff
@@ -48,15 +48,37 @@ impl Default for ProverContextConfig {
...
+            // trade slack for pool without a rebuild.
+            device_slack_static_bytes: env_override(
+                "GPU_PROVER_DEVICE_SLACK_STATIC_BYTES",
+                1 << 27, // 128 MB static slack
+            ),
+            device_slack_per_thread_bytes: env_override(
+                "GPU_PROVER_DEVICE_SLACK_PER_THREAD_BYTES",
+                1 << 11, // 2 KB per thread slack
+            ),
             max_device_allocation_blocks_count: None, // use all available memory
```
robik75: For the combining prover, you can test what the actual allocation high water mark is and use that (with some slack accounting for fragmentation) to determine a value for `max_device_allocation_blocks_count` , that fixes the allocation size and that's it, no need to fudge with slack and as an added bonus by not taking all the VRAM, this prover could run alongside snark wrapper without dropping and recreating instances if the GPU has enough VRAM for both at the same time.

robik75 (changes requested):

full_statement_verifier/src/unified_circuit_statement.rs
```diff
@@ -460,6 +460,22 @@ pub fn verify_unrolled_or_unified_circuit_recursion_layer(
 ) -> [u32; 16] {
     // we just branch
     let op_type = DefaultNonDeterminismSource::read_word();
+    use crate::definitions::*;
+    match op_type {
+        OP_VERIFY_COMBINED_RECURSION_LAYERS_IN_UNIFIED_CIRCUIT => {
+            verify_combined_recursion_layers(security)
+        }
+        _ => verify_single_recursion_layer_op(op_type, security),
```
shamatar: why "match all other"? It anyway requires consistency with what data will be fed from outside, so it should be exact op type, and panic on all other variants

full_statement_verifier/src/unified_circuit_statement.rs
```diff
@@ -480,3 +496,63 @@ pub fn verify_unrolled_or_unified_circuit_recursion_layer(
...
+
+        // Proving chains must be equal.
+        for i in 8..16 {
+            assert_eq!(first_output[i], output[i], "Proving chains must be equal");
+        }
+
+        update_from_recursive_circuit_output(&mut hasher, &output);
+    }
+
+    let mut result = [0u32; 16];
```
shamatar, line 540: Ignoring one of the words in the wrapper circuit is the application level problem (basically wrapper assumes that output is a hash, so truncation is fine). Here there is no such judgement

full_statement_verifier/src/unified_circuit_statement.rs
```diff
@@ -480,3 +496,63 @@ pub fn verify_unrolled_or_unified_circuit_recursion_layer(
...
+
+    result
+}
+
+/// Used in hashing proofs for combining.
+/// Keccak-256, but hashed specifically to be compatible with our SNARK.
+/// First 8 [0 -> 8) words of `output` are the actual output of the circuit, which is what we hash.
+/// Last 8 [8 -> 16) words are the recursion chain, the same across all combined proofs
+/// (checked by the caller) and carried through unchanged.
+fn update_from_recursive_circuit_output(hasher: &mut reduced_keccak::Keccak32, output: &[u32; 16]) {
```
shamatar, line 552: Same here - no shifts or whatever. Ignoring the word is wrapper's problem. So it should be just `hasher.update(&output[0..8])` and nothing else (and for sure not byte by byte)

tools/cli/src/prover_utils.rs
```diff
@@ -636,7 +652,582 @@ pub fn verify_artifact(
             ensure_recursion_chain_binds_program(&output, &unified_level.hash_chain)?;
             Ok(output)
         }
+        ProofTarget::RecursionCombined => {
```
shamatar, line 678: Most likely here changes are needed to be consistent with the verifier changes requested above

shamatar (changes requested):

execution_utils/src/unrolled_gpu.rs
dimazhornyk: I see the concern, but I'd rather address it as a follow-up after this PR to not bloat the scope

full_statement_verifier/src/unified_circuit_statement.rs
dimazhornyk: ok, fixed

full_statement_verifier/src/unified_circuit_statement.rs
dimazhornyk: fixed

gpu_prover/src/prover/context.rs
dimazhornyk: makes sense, I measured  it and set the COMBINER_DEVICE_POOL_BLOCKS to 21_820

full_statement_verifier/src/unified_circuit_statement.rs
dimazhornyk: fixed

gpu_prover/src/prover/context.rs
dimazhornyk: refactored, environment variable override removed

shamatar (dismissed):
Fine on the verifier part

gpu_prover/src/prover/context.rs
```diff
@@ -171,6 +171,12 @@ impl ProverContext {
             let _ = era_cudart::error::get_last_error();
         }
         slack.free()?;
+        log::info!(
+            "GPU device memory pool created: {} bytes ({} blocks of 2^{} bytes) on device {device_id}",
+            arena_bytes,
+            arena_bytes >> allocator_block_log_size,
+            allocator_block_log_size,
+        );
```
robik75, line 179: With this log statement we will get a duplicate log entry because the gpu_worker is also logging nearly the same info.
Either remove or if you want to leave it here, change the log level to debug at most.

robik75 (changes requested):
the new GPU combiner has no GPU CI coverage. The GPU workflow only proves and verifies ordinary recursion-unified artifacts ([workflow lines 100–136](https://github.com/matter-labs/zksync-airbender/blob/1e6b26402d5a2c5bff3083f9d4cbd3c7f0b5211b/.github/workflows/test-gpu.yaml#L100-L136)). Consequently, with_host_state, the fixed 21,820-block arena, GPU combine convergence, and context recreation could regress while CI stays green. Add cli combine --backend gpu plus verification for both security levels; ideally also test two sequential calls through CarriedChainCombiner.

gpu_prover/src/prover/context.rs
dimazhornyk: fixed

robik75 (approved):

PR #354, opened 2026-07-19
refactor(gpu_ntt): deep-clean pass — dead-surface removal, module split, dedup, contract hardening, test trim

robik75:
## What ❔

Deep-clean of `gpu_ntt` (campaign crate 2, following gpu_hash #350). Both NTT engines are production — no engine removal; the win is function-level.

- **Dead-surface removal:** deleted the test-only per-stage / naive-transpose / single-coset launchers + their kernels, and the 12 static `ab_dit_single_*` wrappers (parity tests retargeted to the production stream kernels first).
- **Structure:** split the 1.9k-line `ntt.rs` into per-concern submodules (kernels/inverse/forward/lde/dispatch/shared), demoted 37 internal-only APIs to `pub(crate)`, factored repeated launcher plumbing and CUDA pass-kernel scaffolding into shared helpers.
- **Cross-crate:** hoisted a single guarded `bitreverse_low_bits` into `gpu_core` and collapsed 7 copies/aliases (fixes latent `>>32` UB at `num_bits==0`).
- **Contract hardening:** checked casts, assert-taxonomy pass, and `#[repr(C)]`/`static_assert` layout drift guards mirrored across the Rust↔CUDA boundary.
- **Fixes:** hoisted the LDE domain guards above the fast-path branch (release-mode underflow → silently wrong evals) and a wrong `unreachable!` message.
- **Tests:** trimmed 199 → 141 while adding independent pure-CPU host oracles per compile-time stage-count kernel; reclassified 30 GPU-free tests as `cpu_*`.

Net: 42 files, ~+152 lines (launcher/native reductions offset by new oracles, hardening asserts, and drift guards).

## Why ❔

Aggressive YAGNI + maintainability for the GPU prover stack: remove bloat, make the Rust↔CUDA contract drift-proof, and keep the test suite fast without losing independent coverage of any production kernel.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


gpu/ntt/src/ntt/tests/mod.rs
```diff
@@ -950,27 +783,6 @@ multi_coset_range_parity_test!(
...
-    32,
-    32
-);
-multi_coset_range_parity_test!(
-    multi_coset_monomials_to_evals_log_n_15_cosets_64_base_64,
-    15,
-    12,
-    64,
-    64
-);
```
mcarilli, line 973: IMO it's better to retain these, they're quick to run and help stress the occupancy-tuning logic.

robik75: Good call — restored all three in 4cebff72 (log_n 15/16/17 at cosets 16/32/64). They run in <1ms each under the GPU lock, so no meaningful cost.

For the record on why they're non-vacuous: at these log_n the candidate is the production `lde_with_coset_range` (lde-intermediate / 2pc-batched path) and the oracle is `monomials_to_evals_2_pass_compact_initial` called directly — different kernel families, so it's a real cross-check, not just launch smoke. I also updated the group comment that had called them redundant to note they're kept as occupancy / grid-batching coverage (each log_n is a distinct compiled kernel; correctness is separately pinned by the pure-CPU `host_oracle_*` sets).

PR #365, opened 2026-07-21
feat: EVM verifier

shamatar:
## What ❔

Multiple extensions to support EVM verification of the generated proofs. The compression chain (it'll still happen in Airbender, no other proof systems required) is not yet defined, but the final L1 circuit size is more or less clear already.

## Why ❔

Eliminate all other proof systems for end to end proving all the way to L1

## Is this a breaking change?
- [x] Yes
- [ ] No


verifier_evm/src/templates/gkr.sol
```diff
@@ -0,0 +1,2294 @@
...
+            mstore(GKR_MAIN_GAS_PTR(), gas())
+            ptr, claim,
+            alpha := gkr_compress(ptr, claim, alpha)
+            ptr, claim,
+            alpha := gkr_circuit(ptr, claim, alpha)
+            mstore(GKR_MAIN_GAS_PTR(), sub(mload(GKR_MAIN_GAS_PTR()), gas()))
+        }
+
+        // Permutation identity: read-set product == write-set product (no inversions).
+        check_permutation_identity()
```
yoaveshel, line 713: I think we're missing the lookups checks here (`num == 0  && den != 0`)

shamatar: Hm, indeed

verifier_evm/src/templates/gkr.sol
```diff
@@ -0,0 +1,2294 @@
...
+        // (step1_test/GkrInitYul.t.sol): continue from the post-STEP-1 seed, absorb the 2560 B
+        // of output evals, draw eval_point[4] (→ POINT_PTR) + batching (→ GKR_BATCHING_PTR).
+        // Then compute the 10 dim-reducing claims: eq[16] over eval_point, dotted with each of
+        // the 10 output columns (matches the validated Rust mirror). claim/alpha are legacy
+        // returns kept 0 — gkr_compress reads GKR_CLAIMS_PTR / GKR_BATCHING_PTR.
+        function gkr_init(ptr) -> next_ptr, claim, alpha {
+            let sp := SEED_PTR()
+            // absorb output evals: seed = keccak(seed || outputEvals)
+            mstore(GKR_ABS_PTR(), mload(sp))
+            calldatacopy(add(GKR_ABS_PTR(), 32), ptr, GKR_INIT_BYTES)
```
yoaveshel, line 256: I'm not sure I understand what's happening here, but it seems like we absorb field elements without reducing/asserting <p?

shamatar: Canonical vs non-canonical representation in general is not that important - the knowledge of satisfying witness is still there. We can reduce it, but reductions everywhere would be just expensive. But I'll check separately

verifier_evm/src/templates/gkr.sol
shamatar: Fixed

verifier_evm/src/generator/circuit_yul.rs
```diff
@@ -0,0 +1,2692 @@
...
+            mstore(GKR_ABS_PTR(), mload(SEED_PTR()))
+            let bp := add(GKR_ABS_PTR(), 32)
+{fs_copy}            let s := keccak256(GKR_ABS_PTR(), sub(bp, GKR_ABS_PTR()))
+            mstore(SEED_PTR(), s)
+            s := keccak256(SEED_PTR(), 32)
+            mstore(SEED_PTR(), s)
+            next_alpha := mod(shr(128, s), P)
+            mstore(GKR_ABS_PTR(), mload(SEED_PTR()))
+            bp := add(GKR_ABS_PTR(), 32)
+{ex_copy}            s := keccak256(GKR_ABS_PTR(), sub(bp, GKR_ABS_PTR()))
```
yoaveshel: we batch the evals with `next_alpha` which is drawn before we commit them. Is this fine because they are opening of the whir poly?

verifier_evm/src/templates/whir.sol
```diff
@@ -0,0 +1,636 @@
...
+                sink2(addmod(c0, p1, P), mload(REG_CLAIM))
+                mstore(REG_CLAIM,
+                    addmod(mulmod(addmod(mulmod(c2, alpha, P), c1, P), alpha, P), c0, P))
+                mstore(add(FOLD_CH_PTR, mul(s, 32)), alpha)
+                let zi_off := mload(REG_ZIOFF)
+                fold_accumulator(alpha, mload(REG_NUMPOW), zi_off)
+                mstore(REG_ZIOFF, add(zi_off, 32))
+            }
+
+            switch lt(r, 5)
```
yoaveshel, line 590: Does this mean round 5 is always the final round?

shamatar:
I also ran a separate check to make sure that calldata only contains unique values (no copies of the same logical value at different offsets)

verifier_evm/src/templates/whir.sol
shamatar: Yes, it's derived from WHIR config for L1 mode

verifier_evm/src/generator/circuit_yul.rs
shamatar: Yes, verifier derives evaluation claims itself (unless prover gives evals for cache relations, but we commit them too). The problem is actually in the prover - even though it commits to extra produced evals for caches and so, it does so before drawing the challenge. Will fix too

verifier_evm/src/templates/whir.sol
```diff
@@ -0,0 +1,636 @@
...
+
+        // Every check passed (any failure would have reverted). Notify the registry with
+        // the recomputed committed state — GKR marks the same bytes32 from its own tx.
+        {
+            mstore(0, shl(224, MARK_WHIR_SELECTOR))
+            mstore(4, mload(COMMIT_PTR))
+            pop(call(gas(), REGISTRY, 0, 0, 36, 0, 0))
+        }
+
+        // Return the final claim, the gas used, and the number of proof bytes consumed.
```
yoaveshel, line 609: missing cursor pin (gkr.sol:511 has `if iszero(eq(ptr, sub(calldatasize(), 8))) { revert(0, 0) }`)

shamatar: Do you mean a check that we consumed all the calldata?

yoaveshel: Yes

shamatar: Sure, I can add it, it's cheap

cs/src/gkr_compiler/mod.rs
```diff
@@ -311,24 +270,59 @@ pub enum InitsOrTeardownsTimestampAndValue {
...
+        match (self, other) {
+            (Self::Constant(..), Self::Constant(..)) => {
+                unreachable!("unnormalized")
+            }
+            (Self::Constant(..), _) => std::cmp::Ordering::Less,
+            (_, Self::Constant(..)) => std::cmp::Ordering::Greater,
+            (Self::Place(a), Self::Place(b)) => a.cmp(b),
+            (Self::Place(..), _) => std::cmp::Ordering::Less,
+            (_, Self::Place(..)) => std::cmp::Ordering::Greater,
+            (Self::Product(..), _) => std::cmp::Ordering::Less,
```
yoaveshel, line 307: this isn't a total ordering - `cmp(product_a, product_b)` and `cmp(product_b, product_a)` both resolve to `Less` (same for Sum)

verifier_evm/src/templates/gkr.sol
```diff
@@ -0,0 +1,2329 @@
...
+            for { let i := 0 } lt(i, 8) { i := add(i, 1) } {
+                calldatacopy(add(base, mul(i, 4)), mul(add(18, i), 12), 4)
+            }
+            let setupval := mload(base)
+            // mark_gkr_verified(commitment, public_input, setup_commitment)
+            mstore(0, shl(224, MARK_GKR_SELECTOR))
+            mstore(4, commitment)
+            mstore(36, pubval)
+            mstore(68, setupval)
+            pop(call(gas(), REGISTRY, 0, 0, 100, 0, 0))
```
yoaveshel, line 681: Doesn't `pop` discard the output of `call`, so we never check if the verification actually passed?

verifier_evm/src/templates/gkr.sol
yoaveshel: Doesn't it allow the prover to grind challenges? It can pass `x` or `x+kP` and all arithmetic checks pass but hashes are different

verifier_evm/src/templates/gkr.sol
shamatar: Registry just marks an existence of valid whir/gkr parts for some intermediate state gkr -> whir handoff. We can even do it out of order (whir first, then GKR), and it's external problem to query such register to see if some proof has been fully verified

verifier_evm/src/templates/whir.sol
shamatar: Added

verifier_evm/src/templates/gkr.sol
shamatar: Yes, that's what I want to review

shamatar: I'll remove this degree of freedom

cs/src/gkr_compiler/mod.rs
shamatar: Fixed now too

PR #368, opened 2026-07-23
use stable toolchain + repro fixes

shamatar:
## What ❔

Fix some scripts + make verified binaries to build using stable toolchain


shamatar:
CI for verifier repro should pass now

PR #399, opened 2026-08-18
perf: switch to LSB binding, windowed initial rounds and arch-specific backends

shamatar:
## What ❔

All over performance improvements. NOT: GPU prover is MSB binding, so it's NOT yet compatible. Verifiers were updated to match LSB binding

## Is this a breaking change?
- [x] Yes
- [ ] No


shamatar:
FYI 11.75s on no-charger M3 Max 12 threads for 2^24 add/sub family circuit. 47% of time in sumcheck, 21% initial commit, 32% in WHIR

PR #428, opened 2026-08-31
feat: RAM size dynamic config

shamatar:
## What ❔

We want to allow 4Gb of RAM configurations for at the cost of higher RAM pressure on the simulator machine

## Why ❔

In general 32-bit architecture can address up to 4Gb of RAM

## Is this a breaking change?
- [ ] Yes
- [x] No

### DIFF
--- a/circuit_defs/setups/Cargo.toml
+++ b/circuit_defs/setups/Cargo.toml
@@ -42,7 +42,7 @@ sha3 = { workspace = true }
 hex = "*"
 
 [features]
-defaut = ["witness_eval_fn"]
+default = ["witness_eval_fn"]
 witness_eval_fn = ["add_sub_lui_auipc_mop/witness_eval_fn", "jump_branch_slt/witness_eval_fn", "load_store_subword_only/witness_eval_fn", "load_store_word_only/witness_eval_fn", "mul_div_unsigned/witness_eval_fn", "shift_binary/witness_eval_fn", "unified_reduced_machine/witness_eval_fn"]
 
 [dev-dependencies]
--- a/full_statement_verifier/tests/trace/mod.rs
+++ b/full_statement_verifier/tests/trace/mod.rs
@@ -102,8 +102,11 @@ impl<S: NonDeterminismCSRSource> NonDeterminismCSRSource for StampingSource<S> {
         self.census_marks.push(census_of(&counters));
         self.inner.read()
     }
-    fn write_with_memory_access<R: RamPeek>(&mut self, ram: &R, value: u32) {
+    fn write_with_memory_access<R: RamPeek + ?Sized>(&mut self, ram: &R, value: u32) {
         self.inner.write_with_memory_access(ram, value);
+    }
+    fn write_with_memory_access_raw(&mut self, ram: &[u32], value: u32) {
+        self.inner.write_with_memory_access_raw(ram, value);
     }
     fn write_with_memory_access_dyn(&mut self, ram: &dyn RamPeek, value: u32) {
         self.inner.write_with_memory_access_dyn(ram, value);
--- a/gpu/execution_prover/src/prover/config.rs
+++ b/gpu/execution_prover/src/prover/config.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 
+use riscv_transpiler::jit::JitRunnerRam;
 use riscv_transpiler::vm::SimpleTape;
 use type_map::concurrent::TypeMap;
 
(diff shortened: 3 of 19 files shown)


0xVolosnikov:
### Code review

Found 6 issues (deep correctness pass; JIT offset arithmetic, the `large_timestamps_offset` sequences, `MachineState` field order, and the `process_csr` reconstruction were checked for all four `JitRunnerRam` variants and look correct):

1. `MemoryHolder::allocate_zeroed` allocates with a 2 MiB-aligned `Layout`, but `MemoryHolder` is `repr(transparent)` over `[u64]`, so `Box` drop deallocates with `Layout::for_value` (align 8). Verified with a tracking allocator on the same construction: `ALLOC size=8192 align=2097152` / `DEALLOC size=8192 align=8`. This violates the `Allocator` contract (UB, Miri-visible); benign with glibc `free` today, but a crash with any alignment-aware allocator. The `// Same layout` comment is inaccurate. Suggest a thin owner type with a manual `Drop` that reuses the allocation layout, or dropping the over-alignment.

https://github.com/matter-labs/zksync-airbender/blob/f5219eab99f97be6689b4e49b1a459e8d421ea67/riscv_transpiler/src/jit/structs.rs#L172-L193

2. `MemoryHolder::allocate_uninit` is `todo!()` and is called from `test_perf_with_trace_keeping` (previously a working `Box::new_zeroed`). The test is `#[ignore]` + `skip_if_ci!`, so CI will not catch the panic, but the method is `pub`.

https://github.com/matter-labs/zksync-airbender/blob/f5219eab99f97be6689b4e49b1a459e8d421ea67/riscv_transpiler/src/jit/structs.rs#L166-L170
https://github.com/matter-labs/zksync-airbender/blob/f5219eab99f97be6689b4e49b1a459e8d421ea67/riscv_transpiler/src/jit/tests.rs#L1988-L1991

3. `reset_buffer()` zeroes the whole memory+timestamps buffer on every batch (1.5 GiB at `Medium`, 12 GiB at `Full`), replacing the old zeroing of `[binary_image_len..ROM_WORD_SIZE]`. It appears redundant: `collect_inits_and_teardowns` already zeroes every touched word and timestamp on the normal path, and the abort path calls `MemoryHolder::reset`. The old line is left commented out and the `ROM_WORD_SIZE` import is now unused.

https://github.com/matter-labs/zksync-airbender/blob/f5219eab99f97be6689b4e49b1a459e8d421ea67/gpu/execution_prover/src/workers/simulation_runner.rs#L283-L286

4. `collect_inits_and_teardowns` changed from `&mut MemoryHolder` to `&MemoryHolder`, but the workers still cast `values.as_ptr()` / `timestamps.as_ptr()` (derived from shared `memory()` / `timestamps()` borrows) to `*mut` and write through them. The cast pattern predates this PR, but the exclusive borrow that justified it is gone. Suggest `memory_and_timestamps_mut()` + `split_at_mut`.

https://github.com/matter-labs/zksync-airbender/blob/f5219eab99f97be6689b4e49b1a459e8d421ea67/gpu/execution_prover/src/workers/cpu.rs#L284-L304

5. The `Full` (4 GiB) codegen path has no reference comparison: the only test using `JitRunnerRam::Full` prints state without asserting, while `run_and_compare` / `packed_ts_vs_reference` run at `Medium`. Also `program_prover` still asserts `ram_bound <= 1 << 30`, so `Full` cannot be used for proving yet.

https://github.com/matter-labs/zksync-airbender/blob/f5219eab99f97be6689b4e49b1a459e8d421ea67/riscv_transpiler/src/jit/tests.rs#L484-L493
https://github.com/matter-labs/zksync-airbender/blob/f5219eab99f97be6689b4e49b1a459e8d421ea67/program_prover/src/unified_transition.rs#L169-L173

6. The PR is marked non-breaking, but it removes `jit::RAM_SIZE`, adds a required `ram_config` argument to `run_alternative_simulator` and friends, makes `Context.implementation` crate-private, and adds a required `write_with_memory_access_raw` method to `NonDeterminismCSRSource`. Downstream consumers (e.g. airbender-platform, which imports `RAM_SIZE` and calls `run_alternative_simulator`) will need changes.

Minor: `run` / `run_over_prepared_memory` assert the JIT config against the holder size but not against `Context::ram_config`, which `Context::write_nondeterminism` uses to size the memory slice; the `initial_memory.len() <= ROM_WORD_SIZE` check in `run` was loosened to the whole RAM; the `defaut` -> `default` fix in `circuit_defs/setups/Cargo.toml` newly enables the `witness_eval_fn` default feature for all dependents and is unrelated to this PR.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

